PR #106, opened 2017-11-17
Compilation: Fix custom attributes using types defined in the workbook

### LINKED ISSUE #104: Using types defined in a workbook as attribute parameters doesn't work
#### Xamarin Workbooks
| Detail          | Value                       |
| :-------------- | :-------------------------- |
| Version         | 1.4.0-beta1                 |
| Git Branch      | master                      |
| Git Hash        | 2e30f47                     |
| VSTS Definition | Workbooks & Inspector (Win) |

#### System Info
| Component       | Value                                      |
| :-------------- | :----------------------------------------- |
| Windows         | Microsoft Windows NT 10.0.16299.0 (64-bit) |
| CPU Cores       | 8                                          |
| Physical Memory | 16 GB                                      |

#### System-Installed Software
| Component         | Version    |
| :---------------- | :--------- |
| Xamarin Simulator | 1010103100 |

#### VisualStudio.15.Release Components
| Component                     | Version           |
| :---------------------------- | :---------------- |
| Visual Studio Enterprise 2017 | 15.4.2+27004.2006 |
| Xamarin iOS                   | 11.2.0-0          |
| Xamarin Android               | 8.0.2-1           |

#### VisualStudio.15.Preview Components
| Component                     | Version                |
| :---------------------------- | :--------------------- |
| Visual Studio Enterprise 2017 | 15.5.0-pre.3.0+27102.0 |
| Xamarin iOS                   | 11.4.0-98              |
| Xamarin Android               | 8.1.0-22               |

### Steps to Reproduce

1. Run the workbook from https://gist.github.com/bojanrajkovic/8bf22ea39bb45159de9f2b590a565f38.

### Expected Behavior

The last cell should return 1 instance of the `CustomAttribute` attribute.

### Actual Behavior

`TypeLoadException` is thrown inside the BCL.

### Notes

This smells like a bug I fixed previously with Newtonsoft.Json not being able to load types that were defined in the workbook when deserializing JSON with fully qualified type names, so I'm assigning this to myself to look into it.

bojanrajkovic:
Custom attributes (either from libraries or defined in the workbook itself) would cause the CLR
to throw `TypeLoadException` (with an empty type name!) when trying to load the type to create
the custom attribute instance, somewhere deep inside COM bits (I tracked it to CreateCaObject,
which is an internal call into the CLR).

Changing our assembly name prefix to not have any spaces fixed the initial submission (no
`TypeLoadException`, the correct output from the sample workbook in #104), but subsequent
submissions would throw `TypeLoadException` (this time with a non-bogus type name!) because
the assembly name had changed. 

The 2nd commit in the series fixes that as well, by improving the detection for when the assembly name has ended—the `+` used for nested classes wasn't being detected.

Adds the workbook from #104 as a regression workbook as well--this is a good test to have on hand,
as the kind of scenario this is likely to get hit in (JSON.Net + custom `JsonConverter`) is important.

Fixes #104.

#### To-do

- [x] Smoke test on Mac
  - [x] Console
  - [x] iOS
  - [x] .NET Core
  - [x] Android
  - [x] Mac Mobile
  - [x] Mac Full
- [x] Smoke test on Windows
  - [x] iOS
  - [x] Android
  - [x] WPF
  - [x] Console
  - [x] .NET Core

### DIFF
--- a/Clients/Xamarin.Interactive.Client/Compilation/Roslyn/MonoScriptCompilationPatcher.cs
+++ b/Clients/Xamarin.Interactive.Client/Compilation/Roslyn/MonoScriptCompilationPatcher.cs
@@ -112,7 +112,7 @@ namespace Xamarin.Interactive.Compilation.Roslyn
 
         /// <summary>
         /// Search the target byte array for an 8-byte magic value, then rewrite the string prefixed
-        /// by the magic value. Strings are terminated by '\0', '.', or '>'. This is very aggressive
+        /// by the magic value. Strings are terminated by '\0', '.', '>', or '+'. This is very aggressive
         /// and has nothing to do with the PE format, so the magic value should be very unique!
         /// </summary>
         static unsafe void Patch (byte [] target, byte [] magicBytes, PatchHandler patchHandler)
@@ -131,7 +131,7 @@ namespace Xamarin.Interactive.Compilation.Roslyn
                         var length = 8;
                         while (true) {
                             var b = target [i + length];
-                            if (b == 0 || b == '.' || b == '>')
+                            if (b == 0 || b == '.' || b == '>' || b == '+')
                                 break;
                             length++;
                         }
--- a/Clients/Xamarin.Interactive.Client/Compilation/Roslyn/RoslynCompilationWorkspace.cs
+++ b/Clients/Xamarin.Interactive.Client/Compilation/Roslyn/RoslynCompilationWorkspace.cs
@@ -500,7 +500,7 @@ namespace Xamarin.Interactive.Compilation.Roslyn
             workspace.RemoveProject (project.Id);
         }
 
-        const string assemblyNamePrefix = "🐵 XIS";
+        const string assemblyNamePrefix = "🐵🐻";
         static readonly byte [] assemblyNamePrefixBytes = Encoding.UTF8.GetBytes (assemblyNamePrefix);
 
         ProjectInfo CreateSubmissionProjectInfo ()
--- a/Tests/Workbooks/Regression/Custom Attribute Types.workbook
+++ b/Tests/Workbooks/Regression/Custom Attribute Types.workbook
@@ -1,0 +1,26 @@
+---
+uti: com.xamarin.workbook
+id: 60f1e237-ac68-4dd6-8069-4f1ab3049a28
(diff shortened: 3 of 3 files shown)


bojanrajkovic:
build

bojanrajkovic:
This smoked clean on all platforms, and is mergeable, however, it's technically still broken on platforms that patch the assembly name to include the sequence number (Console/Mono, iOS, Android, Mac).

I want to do a little more investigation, but I need to dump the assemblies generated by the compilation pass to do this. I suspect this has something to do with the way that types are encoded when passed to custom attributes.

Clients/Xamarin.Interactive.Client/Compilation/Roslyn/RoslynCompilationWorkspace.cs
```diff
@@ -398,6 +399,20 @@ public bool HaveAnyLoadDirectiveFilesChanged (DocumentId documentId)
 
             this.includePeImagesInResolution = includePeImagesInResolution;
 
+            var host = ClientApp.SharedInstance.Host;
+            var hostIsWindows = !host.IsMac;
+
+            // On a Windows host, we can skip patching for Console, .NET Core, and WPF, but
+            // must patch for everything else.
+            if (hostIsWindows) {
```
sandyarmstrong: `if (!ClientApp.SharedInstance.Host.IsMac)`

bojanrajkovic: Why did I write it this way 🤔...probably had a different structure the first time.

Clients/Xamarin.Interactive.Client/Compilation/Roslyn/MonoScriptCompilationPatcher.cs
```diff
@@ -131,7 +131,7 @@ static unsafe void Patch (byte [] target, byte [] magicBytes, PatchHandler patch
                         var length = 8;
                         while (true) {
                             var b = target [i + length];
-                            if (b == 0 || b == '.' || b == '>')
+                            if (b == 0 || b == '.' || b == '>' || b == '+')
```
sandyarmstrong: Update the doc comment

PR #113, opened 2017-11-28
Client: Store last used workbook app and select it on startup.

### LINKED ISSUE #58: Store last workbook type for new workbook dialog as preference
The last workbook type created should be the default selection.

bojanrajkovic:
Sets the agent type when either the broader agent grouping or the more specific agent type within the grouping is set and changes constructors to use field setters instead of property setters to avoid side effects as necessary so that we don't update the preference all the time when constructing the items/controller.

Fixes #58.

### DIFF
--- a/Clients/Xamarin.Interactive.Client/Client/ViewControllers/NewWorkbookItem.cs
+++ b/Clients/Xamarin.Interactive.Client/Client/ViewControllers/NewWorkbookItem.cs
@@ -55,7 +55,7 @@ namespace Xamarin.Interactive.Client.ViewControllers
                     true))
                 .ToArray ();
 
-            SelectedWorkbookApp = WorkbookApps [0];
+            selectedWorkbookApp = WorkbookApps [0];
         }
 
         void NotifyPropertyChanged ([CallerMemberName] string propertyName = null)
--- a/Clients/Xamarin.Interactive.Client/Client/ViewControllers/NewWorkbookViewController.cs
+++ b/Clients/Xamarin.Interactive.Client/Client/ViewControllers/NewWorkbookViewController.cs
@@ -5,31 +5,29 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.ComponentModel;
 using System.Linq;
 using System.Runtime.CompilerServices;
 
-using Xamarin.Interactive.Client.AgentProcesses;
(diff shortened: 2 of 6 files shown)


Clients/Xamarin.Interactive.Client/Client/ViewControllers/NewWorkbookViewController.cs
```diff
@@ -53,8 +54,11 @@ public NewWorkbookViewController ()
                 select new NewWorkbookItem (appGroup.First ().Icon, appGroup.Key, appGroup)
             ).ToList ());
 
-            if (Items.Count > 0)
-                SelectedItem = Items [0];
+            var lastUsedAgentType = Prefs.UI.LastUsedAgentType.GetValue ();
+            if (lastUsedAgentType == AgentType.Unknown && Items.Count > 0)
+                selectedItem = Items [0];
+            else
+                SelectedAgentType = lastUsedAgentType;
```
bojanrajkovic: This intentionally uses the property rather than duplicating the logic from the setter.

Clients/Xamarin.Interactive.Client/Preferences/Prefs.cs
```diff
@@ -42,6 +42,9 @@ public static class Theme
                 public static readonly Preference<bool> UseHighContrast
                     = new Preference<bool> ("ui.theme.useHighContrast", false);
             }
+
+            public static readonly Preference<AgentType> LastUsedAgentType
+                = new Preference<AgentType> ("ui.lastUsedAgentType", AgentType.Unknown);
```
sandyarmstrong: I'm wondering if this pref should instead be the workbook app identifier ("console-netcore" for example). Since we are presumably getting rid of `AgentType` when it becomes possible for new workbook apps/types to be added dynamically.

Additionally, does anyone think we should be storing the state of optional features (XF being the only example currently)?

bojanrajkovic: Good point--I keep forgetting that `AgentType` is going away. Not sure about optional features.

Clients/Xamarin.Interactive.Client/Client/ViewControllers/NewWorkbookItem.cs
```diff
@@ -11,6 +11,8 @@
 using System.Linq;
 using System.Runtime.CompilerServices;
 
+using Xamarin.Interactive.Preferences;
```
sandyarmstrong: can go away now?

bojanrajkovic:
build

Issue #23, opened 2017-11-06
Distribute Inspector with IDE extensions instead of in Workbooks installer


sandyarmstrong:
Although Workbooks and Inspector clients are based on the same code, distributing them together is no longer helpful. The Inspector client and necessary agent files should be distributed with the the Inspector extension itself (which is already shipped as part of the Visual Studio Enterprise products), not with the Workbooks installers.

The current plan is to still build the Inspector client and the various agents out of the Workbooks git repo, but the build output for them would change into a NuGet package that the IDE extensions could consume.

Since this would add up to 100MB or so to each IDE's on-disk footprint, we need to get approval from IDE leads.

Additionally, we would want to keep shipping Inspector files in our installers until after the bundled Inspector IDE work hit the stable release channel.

sandyarmstrong:
The work of creating the client NuGet packages is in https://github.com/Microsoft/workbooks/pull/121.

The IDE extensions will be updated separately.

PR #121, opened 2017-12-04
Bundle Inspector clients into NuGet packages for IDE extensions

sandyarmstrong:
This is the first step to fixing https://github.com/Microsoft/workbooks/issues/23.

Both `Xamarin Workbooks.app` and `Xamarin Inspector.app` now have all files they need directly in the app bundle, instead of depending on `/Library/Frameworks/Xamarin.Interactive.framework`. Specifically, forms support and (in the case of Inspector) agent assemblies have been moved to `SharedSupport/Agents`. We can remove `Xamarin.Interactive.framework` as soon as we're ready to drop 15.5 support.

`updateinfo` and `buildinfo` have moved from `Resources` to `SharedSupport` within the app bundles.

Our package build now also generates a `Xamarin.Inspector.Mac` NuGet package, which can be consumed by the VSmac Inspector extension, so that the extension will no longer require a compatible installation of Xamarin Workbooks.

A corresponding `Xamarin.Inspector.Windows` NuGet package for VS has also been added, with no changes to the workbooks client install.

`InteractiveInstallation` has been trimmed down to reflect the fact that we don't need to locate agent files (or clients) from either client app. On Windows and Mac, "installed" items are found relative to the client exe, instead of using absolute framework paths or checking registry keys.

Update checking has been disabled in all Inspector clients.

### DIFF
--- a/Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
+++ b/Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 
 using Foundation;
 using AppKit;
+using ObjCRuntime;
 
 using Xamarin.Interactive.Core;
 using Xamarin.Interactive.I18N;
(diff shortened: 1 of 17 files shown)


sandyarmstrong:
How this will work once we adopt the new package in VSmac:

image

Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
```diff
@@ -96,21 +97,33 @@ public override void WillFinishLaunching (NSNotification notification)
         {
             MenuManager = new MenuManager (NSApplication.SharedApplication.MainMenu);
 
-            ClientApp.SharedInstance.Updater.CheckForUpdatesPeriodicallyInBackground (
-                update => UpdateHandler (null, update));
+            if (ClientInfo.Flavor != ClientFlavor.Inspector)
```
abock: I'm not necessarily against this (and related subsequent checks), but it makes it impossible to update the non-IDE-bundled workbooks/inspector installation.

Clients/Xamarin.Interactive.Client.Windows/Views/MenuManager.cs
```diff
@@ -191,11 +191,13 @@ public void Update (IEditor editor = null)
                 Command = Commands.Commands.Help,
             });
 
-            helpMenu.Items.Add (new MenuItem {
-                Header = "Check for Updates",
-                Command = Commands.Commands.CheckForUpdates,
-                CommandParameter = window
-            });
+            if (ClientInfo.Flavor != ClientFlavor.Inspector) {
```
abock: Unnecessary braces

Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
```diff
@@ -96,21 +97,33 @@ public override void WillFinishLaunching (NSNotification notification)
...
+                    string.Empty);
+                appMenu.InsertItem (checkForUpdatesMenuItem, 1);
+
+                var enableTerminalUsageMenuItem = new NSMenuItem (
+                    Catalog.GetString ("Enable Terminal Usage…"),
+                    new Selector ("installCommandLineTool:"),
+                    string.Empty);
+                appMenu.InsertItem (enableTerminalUsageMenuItem, 2);
+
+
```
abock: Extra blank line

Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
sandyarmstrong: Yeah, I thought about it and decided that was fine. More consistent, too.

sandyarmstrong:
@abock I'm not really sure how we'd say "bundled with IDE x.y.z" without having the IDE extensions pass more info to the client at launch time. And we could do that...I'm just not super sure how useful the extra work would be.

We already know the IDEs they have from Copy Info. We can determine from that if they're using a bundled Inspector. So I don't think it's useful for _us_.

There are probably some Inspector users who will notice the split. But they're probably also the type that read release notes so I just don't think we need to go to the extra effort here (which is why I prefer to hide update UI instead of adding messaging there).

Side note: it will be very easy to change our minds about this later now that we can push new clients to the IDE extensions (mostly) at our whim. :-)

bojanrajkovic:
In terms of finding out which IDE version the client was bundled with, you could walk up the tree and look for the folder containing the VS isolation file (devenv.isolation.ini), which has all of that information, if we wanted it. Looks good otherwise, though!

abock:
@bojanrajkovic yeah, the IDE doesn't need to pass anything. Just walk up.

sandyarmstrong:
Good point on IDE sniffing!

PR #123, opened 2017-12-05
Add "Report an Issue" help menu item

abock:
Review each commit in order.

### DIFF
--- a/Clients/Xamarin.Interactive.Client/IssueReport.cs
+++ b/Clients/Xamarin.Interactive.Client/IssueReport.cs
@@ -10,6 +10,10 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 
+using CommonMark;
+using CommonMark.Formatters;
+using CommonMark.Syntax;
+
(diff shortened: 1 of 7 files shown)


Clients/Xamarin.Interactive.Client/IssueReport.cs
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        public string GetIssueReportForGitHub ()
+        {
+            var writer = new StringWriter ();
+            WriteIssueReportForGitHubAsync (writer).GetAwaiter ().GetResult ();
+            return writer.ToString ().TrimEnd ();
+        }
+
+        public Task WriteIssueReportForGitHubAsync (TextWriter writer)
```
sandyarmstrong: Why is this `async`? Nothing `async` happens in it.

abock: Because originally async stuff was happening (see `WriteEnvironmentMarkdownAsync`); not necessary any more.

PR #133, opened 2017-12-11
Mac: Use paths.d to get workbook CLI tool into PATH.

### LINKED ISSUE #56: Use /etc/paths.d for workbook
We should use `/etc/paths.d` instead of writing out a wrapper script to `/usr/local/bin/workbook`

bojanrajkovic:
Still generates the wrapper script as part of `postinstall`, but instead of writing it directly into `/usr/local/bin`, it's written into our bundle and that path is added to `/etc/paths.d`.

Fixes #56.

### DIFF
--- a/Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
+++ b/Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
@@ -158,34 +158,32 @@ namespace Xamarin.Interactive.Client.Mac
         static async Task InstallCommandLineToolAsync ()
         {
             try {
-                const string scriptPath = "/usr/local/bin/workbook";
-                var authOpen = await Security.AuthOpen.PreauthorizeAsync (
+                const string pathsDPath = "/etc/paths.d/workbooks";
+                var pathsDAuthOpen = await Security.AuthOpen.PreauthorizeAsync (
                     new XamCore.Security.AuthorizationEnvironment {
                         Prompt = Catalog.Format (Catalog.GetString (
-                            "{0} would like to install a script to " +
+                            "{0} would like to install a path helper to " +
                             "support terminal usage: {1}",
                             comment: "{0} is the app name; {1} is a file path"),
                             ClientInfo.FullProductName,
-                            scriptPath),
+                            pathsDPath),
                         AddToSharedCredentialPool = true
                     },
-                    scriptPath,
-                    mode: 493 /*0755*/);
+                    pathsDPath,
+                    mode: 420 /* 0644 */);
 
-                await authOpen.WriteAsync (h => h.WriteData (NSData.FromString (
-                    "#!/bin/bash\n" +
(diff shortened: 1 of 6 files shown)


bojanrajkovic:
Updated to update in-app `Enable Terminal Usage` menu item.

Spent a bit of time working out how to do an elevated _delete_ of the legacy file from the app itself, but I couldn't find a good way that didn't involve shelling out to an external program, etc.— it's probably fine to leave it in the installer only for now, since 1.4 will still have that.

Clients/Xamarin.Interactive.Client.Mac/AppDelegate.cs
```diff
@@ -172,17 +175,35 @@ public void InstallCommandLineTool (NSObject sender)
...
+                    Path.GetDirectoryName (NSBundle.MainBundle.ExecutablePath)
+                )));
+
                 new NSAlert {
                     AlertStyle = NSAlertStyle.Informational,
                     MessageText = Catalog.GetString ("Installation Complete"),
                     InformativeText = Catalog.Format (Catalog.GetString (
                         "The '{0}' tool has been installed to {1}. " +
-                        "Ensure that {1} is in your PATH.",
+                        "Close and reopen your terminal to ensure that it is in PATH.",
```
sandyarmstrong: Since you no longer use `{0}` or `{1}`, should you remove them and the translator comment? Or are you leaving them in case we want to update the string in the future?

bojanrajkovic: Huh? They're both used in the earlier part of the string?

sandyarmstrong: Of course they are. Hold on, sorry, gotta answer that. Nice chatting with you!

sandyarmstrong:
@abock I thought the script is at `/Applications/Xamarin Workbooks.app/Contents/MacOS/workbook` now? I don't see anything going to `/usr/local/bin`.

bojanrajkovic:
Works for me—I actually tried relative paths at first but was having some issues with quoting and decided to go for a minimal set of changes.

abock:
Okay, so you're now generating `Contents/MacOS/workbook`. There are two issues:

1. You don't need to do that now. What issues were you having with the script? You'd base the real path off `$0`.
2. `MacOS` should contain binaries, not scripts; further, this exposes `Xamarin Workbooks` under `PATH`, which is why I suggested `SharedSupport/path-bin`, so we're free to dump more stuff there in the future to be available under `PATH` (but more importantly, to isolate `workbook`).

IOW, the only thing you need to pre-auth is for the writing of `/etc/paths.d/workbooks`.

abock:
```bash
#!/bin/bash

set -e

bundle_macos_dir="$(cd "$(dirname "$0")/../../MacOS" && pwd)"
"$bundle_macos_dir/Xamarin Workbooks" cli "$@
```

bojanrajkovic:
*nod* Good idea on building the real path from `$0`. The shared support stuff makes total sense.

Issue #22, opened 2017-11-06
Fully split individual workbook app support out of the client


sandyarmstrong:
There are a few goals here:

1. Begin defining the API a third party would use to provide their own workbook apps.
2. Break up our build so that identical components (such as the Android app and iOS app) do not have to take up build time on both Mac and Windows.
3. Make it possible to ship a smaller installer, and download large workbooks apps on-demand.
4. Allow us to ship a larger variety of workbook apps without worrying about bloating the installer.

There is a proof-of-concept branch that needs to be updated and migrated to the new repository. It begins the work of bundling workbook apps with their client support assemblies in NuGet packages, which would allow us to use NuGet as our distribution platform for workbook apps.

The proposed layout for such a NuGet package is:

```
/
  xamarin.workbooks/
    app/
      net45/ (typically able to share app everywhere)
      mac/ (use these if necessary)
      win64/
    client-support/ (AgentProcess, coord mapper, etc)
      net45/ (more likely to have these split)
      mac/
      win64/
```

The `workbookapp.json` file would be in `xamarin.workbooks/app/{platform}/`, and would make use of the `appManagerAssembly` property to name the client integration assembly that should be loaded from `xamarin.workbooks/client-support/{platform}/`.

Ideally we would move public API into `Xamarin.Interactive.dll` from `Xamarin.Interactive.Client.dll`, to prevent having multiple assemblies with public API.

PR #135, opened 2017-12-11
Move support for IAgentProcess to Xamarin.Interactive.dll

sandyarmstrong:
The first step of https://github.com/Microsoft/workbooks/issues/22.

Minor refactoring to allow IAgentProcess implementations to drop
their references to `Xamarin.Interactive.Client.dll`.

Notable changes:

* Introduced a simple `ImmutableArray<T>` implementation (similar
  to our existing `ImmutableList<T>` implementation).
* `AgentProcessManager` is no longer generic, and the attribute
  used to register agent processes is now
  `AgentProcessRegistrationAttribute`. Agent processes register
  their own type now. This is a bit cleaner, and lets us avoid
  more refactoring so `AgentProcessManager` can stay in
  `Xamarin.Interactive.Client.dll`.
* `Xamarin.Workbooks.Client.iOS` now has an explicit reference
  on `Newtonsoft.Json` `9.0.1`. We will want to keep this version
  in sync with the version brought in via
  https://www.nuget.org/packages/NuGet.ProjectModel/4.3.0 .


abock:
build

Agents/Xamarin.Interactive/Client/AgentProcesses/IAgentProcessManager.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Xamarin.Interactive.Core;
+
+namespace Xamarin.Interactive.Client.AgentProcesses
+{
+    interface IAgentProcessManager
```
abock: Separate files for each type.

Agents/Xamarin.Interactive/Collections/ImmutableArray.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Xamarin.Interactive.Collections
+{
+    /// <summary>
+    /// An intentionally simple, immutable array. To be replaced via NuGet/ILRepack when
+    /// we can depend on System.Collections.Immutable or some other NuGet.
+    /// </summary>
+    sealed class ImmutableArray<T> : IReadOnlyList<T>
```
abock: Ugh. So what all was using this that you needed to move? I think ultimately I want to just depend on S.C.Immutable (but we'll have to ILRepack).

Agents/Xamarin.Interactive/Collections/ImmutableList.cs
```diff
@@ -59,6 +59,14 @@ public ImmutableList<T> InsertRange (int index, IEnumerable<T> items)
             return new ImmutableList<T> (newItems);
         }
 
+        public int FindLastIndex (Predicate<T> match)
+        {
+            for (var i = items.Count - 1; i >= 0; i--)
```
abock: `for` brace

Agents/Xamarin.Interactive/Collections/ImmutableList.cs
```diff
@@ -59,6 +59,14 @@ public ImmutableList<T> InsertRange (int index, IEnumerable<T> items)
             return new ImmutableList<T> (newItems);
         }
 
+        public int FindLastIndex (Predicate<T> match)
+        {
```
abock: Throw `ArgumentNullException` for `match`

Agents/Xamarin.Interactive/Collections/ImmutableArray.cs
sandyarmstrong: `Message` stuff and `AgentConnection`. And I agree. Fortunately there's very little `Add`ing going on so I think we can just ignore it until we have time to spend on ILRepack.

sandyarmstrong: I may not have needed to move `AgentConnection`. And I might be able to get away with only moving interface stuff for `Message`.

sandyarmstrong: Well, as it stands, we do need at least `Message.cs`. It would be more refactoring to change that. Like adding message creation API to `IMessageService`.

Agents/Xamarin.Interactive/Collections/ImmutableList.cs
sandyarmstrong: Done.

Agents/Xamarin.Interactive/Collections/ImmutableArray.cs
sandyarmstrong: I undid the `AgentConnection` move, but am leaving `Message` stuff alone. Splitting it up seems like it would just be frustrating.

PR #144, opened 2017-12-20
Build: Update sign targets to account for files moving around.

bojanrajkovic:
We never updated sign targets to account for workbook apps moving to _build, and never updated for client integrations or proprietary changes.


abock:
Patch applied in #145

PR #145, opened 2017-12-20
Win: Add deep signature verification for MSI installer

abock:
Keep an exclusion manifest of files that are not signed but that we should not sign either (e.g. open source bits not under our control). Run this tool in VSTS against the final signed installer to ensure everything that should be signed is signed.

Running the verification this way on each merge build will save a ton of time when we bump the release in Visual Studio proper, which also does this verification, but by the time it does so, it's almost surely too late and panic ensues!


abock:
build

abock:
build

dependencies.json
```diff
@@ -144,6 +161,6 @@
   {
     "Kind": "GitSubmodule",
     "Name": "External/Xamarin.PropertyEditing",
-    "Version": "953d37877f78fdf1fa99a33e189aa99a6b5f4655"
+    "Version": "1ad39236886f30dded16811e588ae4accddc4a19"
```
bojanrajkovic: Is this intentional?

lewing: yeah it is correct

abock: Yes, this was missing from @lewing's bump. It reflects the ref on master.

PR #152, opened 2017-12-22
Build: Change where ReleaseVersion_SemVerNuGetSafe gets created

sandyarmstrong:
Previously this was created in `Metadata.targets` along with
`ReleaseVersion_SemVer`, but I failed to realize that this property was
meant to be consumed and then overwritten later in
`Xamarin.Build.targets`.

Moved `ReleaseVersion_SemVerNuGetSafe` to be generated there as well so
it can benefit from `ReleaseVersion` logic that excludes build numbers
when they are not needed. And also because it always should have been
there.


sandyarmstrong:
@abock good point, thanks! Test added.

PR #153, opened 2017-12-22
InstallerVerifier: use WinTrustVerify

### LINKED ISSUE #149: Use managed in-proc signature verification + parallelization for MSI verify
To improve upon #145 there are two big things we can do to make it faster:

1. use a parallel for loop for the sign verification
2. use in-process `X509Certificate.CreateFromSignedFile` (likely) instead of shelling out to `signtool.exe`

### LINKED ISSUE #150: Check all installed Windows SDKs for signtool when verifying MSI
Some of the Windows SDKs may be partial SDKs (for example, Windows 10 SDK installed on Windows 8.1), so they may not have `signtool`.

Instead of only checking 1 SDK and bailing, check all the SDKs if `signtool` isn't in the first one.

abock:
Do signature verification in proces via the WinTrustVerify API using the authenticode policy provider and provider flags that match the results of `signtool verify /pa`.

This makes signature verification go from two minutes to five seconds.

* Fixes #149
* Fixes #150

### DIFF
--- a/Tools/InstallerVerifier/InstallerVerifier.csproj
+++ b/Tools/InstallerVerifier/InstallerVerifier.csproj
@@ -3,6 +3,12 @@
     <OutputType>Exe</OutputType>
     <TargetFramework>net461</TargetFramework>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


abock:
🎄 🍆

PR #171, opened 2018-01-16
Telemetry: remove old events, send new ones

abock:
Use a shared object model for the two telemetry events we are now
tracking using a new server. We only care about app session start with
build and OS info, and agent session start events. Submit these new
models to the new server endpoint.


Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -116,12 +139,6 @@ public void Post (ITelemetryEvent evnt)
 
         async Task PostEventOnceAsync (ITelemetryEvent evnt)
         {
-            if (BuildInfo.IsLocalDebugBuild) {
```
sandyarmstrong: So now we're posting telemetry for local debug builds even if there isn't a local telemetry server running?

abock: No. If we don't have a valid endpoint URI, we won't have an `httpClient` instance, and the `Post` method below silently bails if so.

PR #174, opened 2018-01-19
Property editor improvements.

lewing:
This fixes a couple of corner cases in for our representations in the property editor.

* For some properties we return messages not values, render those.
* The generic cast can fail where our representation has a padded value.  Use Convert.ChangeType instead.


abock:
build

PR #177, opened 2018-01-19
Build.proj: prefer $(MSBuildThisFileDirectory)nuget.exe

abock:
For now, makes it easier to run `/t:Package` without having access to provisionator. We should probably make this more automatic/magic/whatever in a second pass.

### DIFF
--- a/Build.proj
+++ b/Build.proj
@@ -28,6 +28,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <Npm Condition="'$(Npm)' == ''">npm</Npm>
+    <NuGet Condition="'$(NuGet)' == '' And Exists('$(MSBuildThisFileDirectory)nuget.exe')">$(MSBuildThisFileDirectory)nuget.exe</NuGet>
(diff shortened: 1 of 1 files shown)


Build.proj
```diff
@@ -28,6 +28,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <Npm Condition="'$(Npm)' == ''">npm</Npm>
+    <NuGet Condition="'$(NuGet)' == '' And Exists('$(MSBuildThisFileDirectory)nuget.exe')">$(MSBuildThisFileDirectory)nuget.exe</NuGet>
     <NuGet Condition="'$(NuGet)' == '' And Exists('$(Provisionator_ArtifactCachePath)nuget.exe')">$(Provisionator_ArtifactCachePath)nuget.exe</NuGet>
```
sandyarmstrong: Should the order be switched here so we prefer provisionator's nuget.exe?

abock: No. If you manually go and drop your own nuget.exe into the repo, you probably really want to use that one?

abock: (Also note where I would like to do a second pass at making this nicer)

PR #180, opened 2018-01-22
Various XamPub fixes from Friday feedback

abock:


abock:
build

Build/Xamarin.Build/XamPub/MSBuild/GeneratePublicationManifest.cs
```diff
@@ -119,10 +122,15 @@ TFile ProcessFileBase<TFile> (string path, string evergreenName = null)
...
+            // deal with MSBuild converting / to \, multiple //, etc.
+            string MakePublishUri (string relativePath)
+                => Regex.Replace ($"{BasePublishUri}/{relativePath}".Replace ("\\", "/"), @"\/+", "/")
+                    .Replace ("https:/", "https://");
+
+            releaseFile.PublishUri = MakePublishUri (fileName);
 
             if (!string.IsNullOrEmpty (evergreenName))
-                releaseFile.EvergreenUri = $"{BasePublishUri}/{evergreenName}";
+                releaseFile.EvergreenUri = MakePublishUri (evergreenName);
```
bojanrajkovic: The issue this attempts to fix isn't actually happening here, it's happening in the _other_ set of `releaseFile.EvergreenUri`. The code looks like this:

```csharp
releaseFile.EvergreenUri =
                    Path.GetDirectoryName (relativePath) + "/" +
                    updaterItem.Groups ["name"].Value +
                    fileExtension;
```
and should look like this:
```csharp
releaseFile.EvergreenUri =
                    relativePath.Replace (Path.GetFileName (relativePath), string.Empty) +
                    updaterItem.Groups ["name"].Value +
                    fileExtension;
```

Path.GetDirectoryName mangles URIs, unsurprisingly.

bojanrajkovic: In my testing, you don't need this code at all, with the ^ fix applied.

bojanrajkovic: This could also be written using `System.Uri` and the `Segments` property--parse `relativePath` into a URI, remove the last segment, and then graft the evergreen filename on, but that's neither here nor there.

abock:
build

Build/Utility.targets
```diff
@@ -109,9 +109,13 @@ Copyright 2017 Microsoft. All rights reserved.
...
     DependsOnTargets="Xamarin_Build_ReadAllProperties">
-    <PropertyGroup>
-      <UpdateInfoFile Condition="'$(OS)' == 'Windows_NT'">..\Package\Windows\updateinfo</UpdateInfoFile>
-      <UpdateInfoFile Condition="'$(OS)' == 'Unix'">..\Package\Mac\updateinfo</UpdateInfoFile>
+    <PropertyGroup Condition="'$(OS)' == 'Windows_NT'">
+      <ProductName>Xamarin Workbooks &amp; Inspector (Win)</ProductName>
+      <UpdateInfoFile>..\Package\Windows\updateinfo</UpdateInfoFile>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(OS)' == 'Unix'">
+      <ProductName>Xamarin Workbooks &amp; Inspector (Mac)</ProductName>
```
sandyarmstrong: Can we start changing this to just "Xamarin Workbooks" right here, or do we need to coordinate with other people?

We probably want to stop advertising ourselves as "Inspector" everywhere starting with 15.6.

abock: We cannot. Certainly not in this PR.

PR #182, opened 2018-01-22
Win: support building installer without proprietary bits

### LINKED ISSUE #176: Make it possible to build the Windows installer without iOS or Android support
* iOS support requires XVS set up to build iOS on a Mac build host
* Android support is proprietary, but our Windows installer unconditionally refers to bits the proprietary component build produces

This makes testing a complete build from the open-source tree a bit tedious.

abock:
Fixes #176

### DIFF
--- a/Build/Xamarin.Build/MSBuild/Which.cs
+++ b/Build/Xamarin.Build/MSBuild/Which.cs
@@ -1,0 +1,68 @@
+﻿//
+// Author:
+//   Aaron Bockover <abock@xamarin.com>
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Linq;
+using System.IO;
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Xamarin.MSBuild
+{
+    public sealed class Which : Task
+    {
+        [Required]
+        public string Program { get; set; }
+
+        public bool FatalIfNotFound { get; set; } = true;
+
+        public string [] PreferPaths { get; set; }
+
+        [Output]
+        public string FullPath { get; set; }
+
+        public override bool Execute ()
+        {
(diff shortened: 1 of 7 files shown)


Package/Windows/Product.wxs
```diff
@@ -2,6 +2,11 @@
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi"
      xmlns:netfx="http://schemas.microsoft.com/wix/NetFxExtension">
 
+  <?if $(var.InstallerKind) = Proprietary ?>
+    <?define AndroidSupport ?>
+    <?define iOSSupport ?>
```
sandyarmstrong: iOS is not proprietary. Is the goal here just to allow building the installer without having a Mac build host set up?

A comment might be helpful here in the future, if that's the case. Or maybe we could decide whether or not to include iOS support separately, so that somebody with proprietary bits could build an installer _with_ Android and _without_ iOS?

abock: Oh, hmm. Good point. I can split that out into a separate property I guess and check explicitly for the app having been built.

abock: But yes - the goal is to allow the installer to build without having built some components that otherwise will build as part of our official VSTS builds.

Build/Xamarin.Build/MSBuild/Which.cs
```diff
@@ -0,0 +1,67 @@
...
+        [Output]
+        public string FullPath { get; set; }
+
+        public override bool Execute ()
+        {
+            var preferPaths = PreferPaths ?? Array.Empty<string> ();
+
+            var extensions = new [] { null, ".exe" };
+            var searchPaths = preferPaths
+                .Concat (Environment.GetEnvironmentVariable ("PATH").Split (';'));
```
sandyarmstrong: Don't you need to split on `:` for Mac?

abock:
build

PR #194, opened 2018-02-13
Move Xamarin.Interactive.Client to netstandard2.0

abock:
Use a forked version of `NuGet.Client` for now which has provides `NuGet.PackageManagement` under `netstandard2.0`. This was the last dependency that was preventing our client library from being `netstandard2.0`. The fork should merge into master soon, but for now take the dependency from our myget.org feed.

* NuGet/NuGet.Client#2011

Split a small amount of code out of `Xamarin.Interactive.Client` into `Xamarin.Interactive.Client.Desktop`, which must still be `net461` (anything depending on XCB and Proppy). Make the Mac and Windows clients reference the new `net461` `Xamarin.Interactive.Client.Desktop`.

Refactor our use of `RoutedUICommand` such that the core client assembly only needs `ICommand` and instantiates via a simple factory.

Minor other random changes are needed as well to build under `netstandard2.0`.

_This is the first big step foward to being able to run fully inside ASP.NET Core 2.0._


sandyarmstrong:
🏂

PR #200, opened 2018-02-15
Bump Roslyn to 2.6.1, enable C# 7.2

abock:


Clients/Xamarin.Interactive.Client.Mac/Xamarin.Interactive.Client.Mac.csproj
```diff
@@ -94,7 +94,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp">
-      <Version>2.3.0</Version>
+      <Version>2.8.0</Version>
```
sandyarmstrong: Any idea what feed 2.8.0 is coming from? It's not on nuget.org. I feel like there should be a way we can track this in dependencies.json so we don't surprise ourselves.

sandyarmstrong: It's a weird version number, too. I wonder if it's a nightly build more than a preview release. I don't see any tags for 2.8.0.

sandyarmstrong: For the record, the latest on nuget.org is 2.6.1.

abock: Ugh, VS Mac. I had selected the nuget.org version and then I guess it upgraded it "for free" from the dotnet feed. And yeah, if only we could track the feed. This is something NuGet itself doesn't seem to be able to do (recall our cache discussion from yesterday).

abock:
Yeah, this should have been 2.6.1. Fixed. VSM upgraded it to the dotnet feed I guess, even though I selected in the UI from nuget.org feed.

PR #218, opened 2018-03-19
Various fixes and improvements for console client

abock:


abock:
Why would there be a deadlock there? FWIW, I tested web and desktop too. Also, this code literally brings back the same functionality we had for a long time over in the old repo: https://github.com/xamarin/inspector/commit/478d81272b30b12ee2c087f02a68cb2eb0188b96#diff-6c12b73c7ed95a02b936ab5e44f08bfd

sandyarmstrong:
:+1: I just instinctively worry when I see a main thread async call that "blocks" on (potentially async) event handlers. If it works, it works!

Issue #216, opened 2018-03-16
Core Build TODO


abock:
Tracking issue for splitting up the build (aka CoreBuild.proj) and removing Build.proj.

- [ ] Tests
- [ ] XVS Build Host Configuration
- [ ] Provisionator
- [ ] Windows installer
- [ ] NuGet: pack Inspector client
- [ ] NuGet: pack agent SDK
- [ ] Windows: `_KillWrenchAdbServer` before build
- [ ] Mac: `xcrun -k` before build
- [ ] Proprietary
- [ ] PDB Files
- [ ] Sign Files
- [ ] `UpdatePublicApiDocs`
- [ ] `AssemblyPublicApiDocs`
- [ ] `UpdatePublicApiDefinitions`

PR #219, opened 2018-03-20
Split Roslyn support into a separate assembly, refine public CodeAnalysis API

abock:
This creates two new assemblies:

* `Xamarin.Interactive.CodeAnalysis` - split out from `Xamarin.Interactive.Client`
   - no longer depends on `Microsoft.CodeAnalysis` (Roslyn)
* `Xamarin.Interactive.CodeAnalysis.Roslyn` - split out from `Xamarin.Interactive.Client`
   - depends on `Microsoft.CodeAnalysis` (Roslyn)
   - does not need to depend on `Xamarin.Interactive.Client`

This means that the core is now 100% Roslyn agnostic. This means we can now:

* Cleanly prototype F# support!
* Cleanly prototype a new Roslyn provider that is based on the MSBuild workspace and newer APIs that were not available two years ago when we initially implemented Roslyn support.

Additionally, all DTO/Model types in `Xamarin.Interactive.CodeAnalysis` are renamed and shaped to mirror types directly in Monaco and are fixed up to be immutable and easily serializable. All intermediate/old view-models have been eliminated as well.

`CoreBuild.proj` is also improved such that it can run either under Desktop (.NET Framework/VS/Mono) `msbuild` or .NET Core's `msbuild`. The following build environments now work and are tested in CI:

* CoreBuild (VS msbuild): VSTS Hosted Windows
   - `msbuild CoreBuild.proj`
* CoreBuild (VS msbuild): VSTS Hosted Mac
   - `msbuild CoreBuild.proj`
* CoreBuild (.NET Core msbuild): Travis Linux
   - `dotnet build CoreBuild.proj`
* OldBuild: VSTS RE Mac
   - `msbuild Build.proj`
* OldBuild: VSTS RE Windows
   - `msbuild Build.proj`
* OldBuild: AppVeyor
   - `msbuild Build.proj`

`CoreBuild.proj` now has more coverage and will build tests. There is still more to port from the old `Build.proj`, which is tracked in #216.

Finally, `Xamarin.Build` is now a `netstandard2.0` SDK style project and is no longer bootstrapped as part of either `Build.proj` or `CoreBuild.proj`. The assembly is stored in the repo (not super happy about this, but it drastically reduces convoluted classic msbuild vs .NET Core msbuild bootstrapping issues) so no bootstrapping is required. It is ILRepack'd so that only a single assembly needs updating. To update, just run `dotnet build` in the `build/Xamarin.Build` directory.


Build.proj
```diff
@@ -142,7 +142,7 @@
       BuildInParallel="true"
       Projects="$(SolutionToBuild)"
       Properties="$(SolutionProperties);$(MSBuildCommonProperties)"
-      Targets="$(InteractiveBuildTarget)"/>
+      Targets="Restore;$(InteractiveBuildTarget)"/>
```
sandyarmstrong: Here and below shouldn't have `Restore`, because then you'd be restoring on `Clean`, too. You should it in the `Build` target instead, right? Where we have `Properties="InteractiveBuildTarget=Build"`?

sandyarmstrong: Lol nevermind, reading next commit now :-P

abock: No, this is fine. Clean goes away in the next commit. Also keep in mind the real goal is nuking Build.proj altogether, so for right now I don't care if it's a little exhaustive, as long as it works in CI.

build/Clean.targets
```diff
@@ -0,0 +1,29 @@
...
+    </ItemGroup>
+
+    <PropertyGroup>
+      <FindRoot>$(TopDirectory.TrimEnd('/').TrimEnd('\\'))</FindRoot>
+      <FindArguments>@(FindDirectoryNamesToRemove -> '-name %(Identity)', ' -or ')</FindArguments>
+    </PropertyGroup>
+    <Exec
+      ConsoleToMSBuild="true"
+      StandardOutputImportance="low"
+      Command="find '$(FindRoot)' -type d -and '(' $(FindArguments) ')'">
```
sandyarmstrong: Don't we need a Windows equivalent of this? Leaving note because we don't test `Clean` in the build and don't want it to be forgotten.

abock: Yes

sandyarmstrong:
I wouldn't change _a thing_.

PR #221, opened 2018-03-22
Minor fix on text formatting (proposal)

kant:


abock:
UPDATES.md is the source for our release notes and conforms to Xamarin documentation style where **this style** is used for user interface navigation/descriptions and not `this style`.

PR #224, opened 2018-03-26
Web: Load workbook packages from Gist and disk.

bojanrajkovic:


Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        const hasMatchingUti = (!rawMetadata.uti || rawMetadata.uti.localeCompare(defaultUti, undefined, { sensitivity: "base" }));
+        if (checkUti && hasMatchingUti)
+            throw new Error("Invalid manifest");
+        else
+            delete rawMetadata.uti
+
+        let id = rawMetadata.id;
+        if (id && isUUID.v4(id))
```
abock: I think it's sufficient to simply check that it's just not null/empty. We definitely don't care what version the GUID may be, or even if it's a GUID. The only reason it's a v4 GUID is because that's all that `System.Guid` will produce via `NewGuid`. We don't even use it anywhere<sup>1</sup>. You can eliminate the `is-uuid` dependency.

_<sup>1.</sup> This originated on the first pass of pages where it was going to be used for referencing. If we ever get back to that, it may not even survive._

Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
```diff
@@ -0,0 +1,190 @@
...
+    "macnet45": "mac-xamarinmac-full",
+    "macmobile": "mac-xamarinmac-modern",
+    "android": "android-xamarinandroid",
+    "wpf": "wpf",
+    "console": "console",
+    "dotnetcore": "console-netcore"
+});
+
+function getWorkbookTargetIdFromPlatform(platform: string): string {
+    return manifestPlatformToWorkbookTargetMap.get(platform.toLocaleLowerCase())
```
abock: These are not locale-specific strings (they're effectively enums). Just use `toLowerCase`.

Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
```diff
@@ -0,0 +1,190 @@
...
+        saveableManifest.packages = this.manifest.packages.map(pd => {
+            return !pd ? undefined : { id: pd.packageId, version: pd.identityVersion }
+        }).filter((_: any) => _).toArray()
+        saveableManifest.platforms = this.manifest.platforms.map(wt => {
+            return !wt ? undefined : getPlatformFromWorkbookTargetId(wt.id)
+        }).filter((_: any) => _).toArray()
+        return saveableManifest
+    }
+
+    parseManifest(rawMetadata: any, workbookSession: WorkbookSession): WorkbookManifest {
```
abock: Would be good to explain the general idea of how this whole method works. It's a little hard to follow with all of the deleting properties out of `rawMetadata`. Should also explain that it will round-trip any and all arbitrary YAML and just pulls out bits that we actually need.

Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
```diff
@@ -0,0 +1,190 @@
...
+        const platform: string = rawMetadata.platform;
+        const platforms: string[] = rawMetadata.platforms || [];
+        finalMetadata.platforms = List(platforms.concat(platform).filter(_ => _).map(workbookPlatform => {
+            const matchingTarget = workbookSession.availableWorkbookTargets.find(value => {
+                return value.id === getWorkbookTargetIdFromPlatform(workbookPlatform)
+            })
+            if (!matchingTarget)
+                console.warn(`Could not find matching target for workbook platform ${workbookPlatform}.`)
+            return matchingTarget;
+        }).filter(_ => _))
```
abock: This call to `.filter` seems entirely pointless. Explain?

Clients/Xamarin.Interactive.Client.Web/ClientApp/components/WorkbookShell.tsx
```diff
@@ -138,26 +140,61 @@ export class WorkbookShell extends React.Component<any, WorkbookShellState> {
         this.fileButton.click();
     }
 
-    loadWorkbook(event: React.ChangeEvent<HTMLInputElement>) {
+    async triggerGistPicker() {
+        if (!this.workbookEditor)
+            return;
+
+        const gistUrl = prompt("Enter the URL of the Gist you want to load from?");
```
abock: `// TODO: real UI` Hello 1996!

Clients/Xamarin.Interactive.Client.Web/ClientApp/components/WorkbookShell.tsx
```diff
@@ -138,26 +140,61 @@ export class WorkbookShell extends React.Component<any, WorkbookShellState> {
...
         const file = event.target.files[0]
         const reader = new FileReader
-        reader.addEventListener('load', () => this.restorePackages (reader))
-        reader.readAsText(file)
+        reader.addEventListener('load', () => this.loadWorkbook(file, reader))
+        reader.readAsArrayBuffer(file)
+    }
+
+    sniffHeader(file: ArrayBuffer): Boolean {
+        const magic = new Uint32Array(file.slice(0, 4))[0]
```
abock: Bounds check? What if 3 byte file?

Clients/Xamarin.Interactive.Client.Web/ClientApp/typings/custom.d.ts
```diff
@@ -0,0 +1,29 @@
+declare module "draft-js-plugins-editor";
+declare module "draft-js-markdown-plugin";
+declare module "turndown";
+declare module "is-uuid";
```
abock: Nuke.

Clients/Xamarin.Interactive.Client.Web/Controllers/GistController.cs
```diff
@@ -0,0 +1,42 @@
+//
```
abock: This entire class needs formatting per our .editorconfig.

Clients/Xamarin.Interactive.Client.Web/package.json
```diff
@@ -47,6 +48,8 @@
     "file-saver": "^1.3.3",
     "gray-matter": "^3.1.1",
     "immutable": "^3.8.2",
+    "is-uuid": "^1.0.2",
```
abock: Nuke

Clients/Xamarin.Interactive.Client.Web/Controllers/GistController.cs
```diff
@@ -0,0 +1,42 @@
...
+                var gist = JObject.Parse(await gistApiResponse.Content.ReadAsStringAsync());
+                var latestGistRevision = (string)gist["history"][0]["version"];
+                var gistOwner = (string)gist["owner"]["login"];
+
+                var gistZipUrl = $"https://gist.github.com/{gistOwner}/{id}/archive/{latestGistRevision}.zip";
+                var gistZipResponse = await hc.GetAsync(gistZipUrl);
+                if (!gistZipResponse.IsSuccessStatusCode)
+                    return StatusCode((int)gistZipResponse.StatusCode);
+
+                return File(await gistZipResponse.Content.ReadAsStreamAsync(), "application/zip", $"{latestGistRevision}.zip");
```
abock: Wrap 3 lines:

```csharp
return File (
    await gistZipResponse.Content.ReadAsStreamAsync (),
    "application/zip",
    $"{latestGistRevision}.zip");
```

Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
bojanrajkovic: 👍 Looks like bad autocomplete from VS Code. Never even _seen_ that method used.

Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
bojanrajkovic: It's filtering out undefined values from the final list--if the rest of the code can deal with undefined values in the workbook targets, that can be removed.

Clients/Xamarin.Interactive.Client.Web/ClientApp/components/WorkbookShell.tsx
bojanrajkovic: 👍 I had tested this and Uint32Array didn't seem to care, but I tested wrong, it does.

Clients/Xamarin.Interactive.Client.Web/Controllers/GistController.cs
bojanrajkovic: Weird, I have the EditorConfig plugin for VSCode...

Clients/Xamarin.Interactive.Client.Web/ClientApp/Workbook.ts
```diff
@@ -0,0 +1,199 @@
...
+        // doesn't know about, without duplicating any of it. We'll remove properties we actually consume
+        // from the `rawMetadata` object (all of the deletes below) so that we can store it as the
+        // preserved metadata.
+        if (!rawMetadata) {
+            rawMetadata = {}
+            checkUti = false
+        } else
+            rawMetadata = Object.assign({}, rawMetadata)
+
+        const hasMatchingUti = (!rawMetadata.uti || rawMetadata.uti.localeCompare(defaultUti, undefined, { sensitivity: "base" }));
```
abock:
```javascript
const hasMatchingUti = !rawMetadata.uti || rawMetadata.uti.toLower() === defaultUti
```

PR #232, opened 2018-04-02
Finish the new build setup; persist properties for subsequent runs

abock:
Notably, write a bunch of properties to disk to load on subsequent builds, including resolved CLI tool paths and the `Configuration` and `Profile` properties. This allows for:

1. No need to specify `Configuration` and `Profile` on subsequent build invocations:
    ```bash
    $ msbuild /p:Configuration=Release /p:Profile=Web
    ...
    $ msbuild # on this invocation, Configuration and Profile will be the same as above

2. Eliminates the need for multiple projects in the build to have to re-resolve/compute (expensive) properties. We can take advantage of this facility in future commits to help speed up the build in some areas and make it less verbose as well.

Further, all global properties and targets are now following the new `Directory.Build.props` and `Directory.Build.targets` model introduced in the new MSBuild so each project does not need to manually import (with what were previously subtle ordering differences).

Finally, all supplementary (e.g. CI step) targets have been unwinded from the dependency graph (e.g. no `DependsOnTargets` or `Before/AfterTargets` coupling) of the core build targets (`Restore`, `Build`, `Package`).

As all supplementary processes have now been ported to `CoreBuild.proj`, this PR renames it back to `build.proj`, and we now have only a single project and no solution in the root of the repo, meaning we do not have to specify a project or solution to build!

```
msbuild
```

```
dotnet build
```


Clients/Xamarin.Interactive.Client/ClientApp/WebApp.targets
```diff
@@ -15,8 +15,7 @@
   </PropertyGroup>
 
   <Target
-    Name="_Browserify"
-    DependsOnTargets="ResolveNodeCli">
+    Name="_Browserify">
     <Exec WorkingDirectory="$(MSBuildThisFileDirectory)" Command="&quot;$(Npm)&quot; install --save-dev" />
```
sandyarmstrong: s/Npm/Yarn now?

abock: No. I am not touching our old stuff to make it work with yarn.

build/ProductAssemblyInfo.cs
```diff
@@ -9,14 +9,16 @@
...
     "Copyright 2014-2016 Xamarin Inc. All rights reserved.")]
 [assembly: AssemblyVersion ("1.0.0.0")]
 
 // Spells 0.A.F.V, a magic version that UpdateBuildInfo will
 // replace with a real file version (see BuildInfoTask).
 [assembly: AssemblyFileVersion ("0.65.70.86")]
 
+#if !PRODUCT_ASSEMBLY_INFO_OMIT_IVT
+
 #if PCL_FACADE_ASSEMBLY
```
sandyarmstrong: the PCL_FACADE_ASSEMBLY section can go away, btw

abock: Ah, right. Nice. Also for future reference: when the rest of our non-SDK-style projects can become SDK-style projects, this whole file can go away and be handled entirely via MSBuild within `Directory.Build.props`. It's magic.

PR #233, opened 2018-04-09
Build: minor polish, update README

abock:


abock:
Unfortunately YAML builds in VSTS do not produce build badges yet.

PR #235, opened 2018-04-11
Build: ILRepack Xamarin.Interactive.dll, reference Newtonsoft.Json

abock:
Agent serialization will be moving from `BinaryFormatter` to JSON via Newtonsoft.Json to keep consistency with the new web stuff. This will also allow communication directly between an "agent" and something that is _not_ .NET (e.g. no `BinaryFormatter` on one end).

This implements a .NET Standard-aware `ILRepack` MSBuild task that also can fully rename internal types to avoid collisions due to `InternalsVisibleTo` (internalizing is not enough - we need full renames).

_Best reviewed commit-by-commit._


.gitignore
```diff
@@ -5,7 +5,8 @@ node_modules/
 packages/
 workbooks-proprietary/
 
-Build/Local.props
```
sandyarmstrong: wasn't local.props where you can store info about your preferred Mac build server?

build/Xamarin.Build/MSBuild/ILRepack.cs
```diff
@@ -0,0 +1,187 @@
...
+            };
+
+            var renames = new List<(Regex regex, string replacement)> ();
+            foreach (var rename in NamespaceRenames)
+                renames.Add ((new Regex (rename.ItemSpec), rename.GetMetadata ("Replacement")));
+
+            var typesToRenamespace = new HashSet<string> ();
+            var renamespaceMap = new Dictionary<string, string> ();
+
+            for (int i = 1; i < inputAssemblies.Count; i++) {
```
sandyarmstrong: Just checking: we skip the first input assembly because it's the "main" one? X.I.dll for example? And therefore it doesn't need any types renamespaced?

abock: Yes. This convention brought to you by ILRepack.

.gitignore
abock: Yes. But it doesn't exist anymore.

sandyarmstrong: Did we want to bring it back in the new build world? I can file an issue if you want. The README still has info about it.

PR #244, opened 2018-04-27
Workbooks on WebAssembly

bojanrajkovic:
## Workbooks on WebAssembly

A workbook app running in the browser using the Mono WebAssembly SDK. Most things are supported (NuGet, agent globals, etc.). Agent-wise, there is currently no support for integrations and aborting evaluations, and generally no support for networking in the WASM agent (this is a WASM limitation).

### Major things left to do (this is an incomplete list):

- [x] The WASM UX
  -  [x] There's no indication that the WASM agent is still not up yet from the UI. In theory, this is fine, because any compilations will be queued for evaluation once the agent comes up, but we should still have some indication that something is indeed going on, and disallow evaluations being submitted until the agent is actually up.
      - This is extra relevant in debug mode on Chrome, where due to [upstream bugs][crbug811373], WASM takes 20-30 seconds to load.
  - [x] Evaluations look unfinished for now because they're missing evaluation state, I think. Need to review the recent exception-handling improvements to understand how to finish an evaluation.
  - [x] Evaluations with errors don't properly get marked as finished, and can't be re-evaluated.
  - [x] No globals at all. EvaluationGlobalStateObject needs an `Agent,` we don't actually have one here, though we potentially could as long as we never call `Start`...
  - [x] Loading the WASM agent really destroys performance/behavior of pasting things, especially into Monaco. It seems to be related to Draft's use of a polyfill for `setImmediate`, which relies on `window.postMessage`, which _happens_ to be what we use for client <-> WASM communication (the WASM "agent" currently runs in an `<iframe>`). Our listener doesn't touch Draft's messages, so I'm not really sure why this happens, but it does.
  - [x] Disable WASM on Safari. Unfortunately, Mono WASM crashes Safari's JS JIT, per @kumpera.
- [x] The rest of the agent features—currently only the most basic evaluations will work. `#r` is definitely a no-go, as is NuGet (it's not disabled in the UI, but it will not work). The agent will receive the list of resolved assemblies but it has no idea how to actually get them for loading. We'll likely want some sort of OOB API for it to ask for assemblies so that we don't have to pass around _more_ Base64-encoded PE images, if possible.
  - [x] Add a whitelist of loaded assemblies via compilations so that the out-of-band assembly loading API doesn't get arbitrary file reads. With this, the server controls which files are allowed to be accessed via the session hub observing compilations and controlling the whitelist.
- [x] The behind-the-scenes implementation is currently hardcoded to issue WASM's special tickets when the WASM target is selected. This doesn't affect other workbook apps, but will need to eventually go through a more generic mechanism via the workbook manifest, as we discussed.
- [x] The SDK download is a custom drop that I built and uploaded to my own server while I work on upstreaming the few patches that we carry on the WASM SDK. The server is generally very well available, but this is not a sustainable long-term strategy. We'll need to get patches upstreamed and/or make a build lane that we can do drops from for patches we can't upstream.
- [x] Make it work from `dotnet build` on !Windows. Right now because it targets `net461` it won't work.

### Build instructions

WebAssembly is integrated into the build.

1. Build the web profile by running `msbuild /p:Profile=Web`
2. Run the web client: `pushd Clients/Xamarin.Interactive.Client.Web/ && ASPNETCORE_ENVIRONMENT=development dotnet run`
3. Navigate to http://localhost:5000 in your browser
4. WebAssembly should be available in the dropdown—switch to it and wait for the loading spinner to go away, you should then be able to evaluate things.

[crbug811373]: https://bugs.chromium.org/p/chromium/issues/detail?id=811373&q=component%3ABlink%3EJavaScript%3EWebAssembly%20&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified


ghuntley:
Paging @arggrande this is probs will be the branch we will be cloning down and hacking on at the end of April for personal development. If we start research now then we will be in a better position to PR @abock.

Agents/Xamarin.Interactive/CodeAnalysis/Evaluating/EvaluationAssemblyContext.cs
```diff
@@ -71,26 +71,30 @@ public void Dispose ()
...
                 if (args.RequestingAssembly?.ReflectionOnly == true) {
                     if (File.Exists (assembly.Content.Location))
                         return Assembly.ReflectionOnlyLoadFrom (assembly.Content.Location);
 
                     if (assembly.Content.PEImage != null)
                         return Assembly.ReflectionOnlyLoad (assembly.Content.PEImage);
 
-                    throw new Exception (
-                        $"Could not reflection-only load assembly {args.Name}, location " +
+                    Log.Verbose (
```
bojanrajkovic: This is changed from an exception because this exception frequently gets eaten up anyway by the assembly load exception that is thrown if/when we return null. This way, we're actually more likely to see this.

Agents/Xamarin.Interactive/CodeAnalysis/Evaluating/EvaluationContextId.cs
```diff
@@ -19,7 +19,7 @@ internal static EvaluationContextId Create ()
             => this.id = id;
 
         public bool Equals (EvaluationContextId id)
-            => id.id == this.id;
+            => this.id.Equals (id.id);
```
bojanrajkovic: Reminder: this change is to work around a bug in WASM.

Agents/Xamarin.Interactive/CodeAnalysis/Evaluating/EvaluationContextManager.cs
```diff
@@ -65,8 +65,9 @@ public void Dispose ()
         readonly Observable<ICodeCellEvent> events = new Observable<ICodeCellEvent> ();
         public IObservable<ICodeCellEvent> Events => events;
 
-        public IAgentSynchronizationContext SynchronizationContexts { get; }
-            = new AgentSynchronizationContext ();
+        readonly Lazy<AgentSynchronizationContext> agentSynchronizationContext
+            = new Lazy<AgentSynchronizationContext> ();
+        public virtual IAgentSynchronizationContext SynchronizationContexts => agentSynchronizationContext.Value;
```
bojanrajkovic: This lets the WASM ECM control this during actual evaluations—there's no sync context in WASM, so trying to use the default value causes it to throw up. In brief testing, using the single-thread sync context that we use on the Console app causes lockups, so I preferred to just leave this null (see `WebAssemblyEvaluationContextManager` in the workbook app).

Clients/Xamarin.Interactive.Client.Web/ClientApp/WorkbookSession.ts
```diff
@@ -124,6 +144,78 @@ export class WorkbookSession {
...
+        this.statusUIActionEvent.dispatch({
+            action: StatusUIAction.DisplayMessage,
+            message: {
+                kind: MessageKind.Status,
+                severity: MessageSeverity.Info,
+                text: "Initializing WebAssembly agent…",
+                showSpinner: true
+            }
+        });
+        this.wasmFrame!.contentWindow!.postMessage({ $type: "ChannelOpenMessage" }, location.origin, [
```
bojanrajkovic: We need to transfer the other port of the message channel to the WASM frame, so we need this one `postMessage` call—everything afterwards goes through the message channel, which should be more efficient, and doesn't mess with Draft/Monaco.

Clients/Xamarin.Interactive.Client.Web/Hubs/InteractiveSessionHub.cs
```diff
@@ -80,22 +84,31 @@ public IObservable<InteractiveSessionEvent> ObserveSessionEvents ()
         {
             var events = GetSession ().Events;
 
-            if (hostingEnvironment.IsDevelopment ()) {
-                var settings = new Serialization.ExternalInteractiveJsonSerializerSettings ();
-                events.Subscribe (new Observer<InteractiveSessionEvent> (evnt => {
+            events.Subscribe (new Observer<InteractiveSessionEvent> (evnt => {
```
bojanrajkovic: It seemed silly to have two different subscriptions, so I moved the `if` check inside this one.

Clients/Xamarin.Interactive.Client.Web/Middleware/WasmStaticFileMiddleware.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Xamarin.Interactive.Client.Web
+{
+    public static class WasmStaticFileMiddleware
+    {
+        public static IApplicationBuilder UseWasmStaticFiles (this IApplicationBuilder builder)
+        {
+            var provider = new FileExtensionContentTypeProvider ();
+
+            provider.Mappings.Add (".wasm", "application/wasm");
```
bojanrajkovic: This mapping is required, otherwise most browsers won't use the file at all.

WorkbookApps/Xamarin.Workbooks.WebAssembly/Client/runtime.ts
```diff
@@ -0,0 +1,498 @@
...
+
+        return new MonoMethod(klass.assembly, klass, name, argc, method, this)
+    }
+
+    /**
+     * Boxes a number so it can be passed to Mono. Must be freed by the caller!
+     * @param number The number to box
+     * @returns A pointer to some memory with the integer boxed into it.
+     */
+    boxInteger(number: number): number {
```
bojanrajkovic: Investigate whether I can get rid of this and boxBoolean below—I think they are not needed anymore.

Agents/Xamarin.Interactive/Serialization/InteractiveJsonSerializerSettings.cs
```diff
@@ -16,6 +16,7 @@
 using Xamarin.Interactive.CodeAnalysis;
 using Xamarin.Interactive.CodeAnalysis.Evaluating;
 using Xamarin.Interactive.CodeAnalysis.Models;
+using Xamarin.Interactive.CodeAnalysis.Resolving;
```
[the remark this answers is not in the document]
bojanrajkovic: Yep, thanks VSCode for not showing me unused namespaces.

Clients/Xamarin.Interactive.Client.Web/ClientApp/WorkbookSession.ts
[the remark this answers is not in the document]
bojanrajkovic: Yeah, I know—this was intended to be a "shut up, I know they're there" signal to TS, as this method is a callback from the frame's load event and it would be extremely confusing if it didn't exist at this point. It'll be a lot cleaner once WASM bits are refactored into a separate object.

WorkbookApps/Xamarin.Workbooks.WebAssembly/Xamarin.Workbooks.WebAssembly.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+      <OutputType>Library</OutputType>
```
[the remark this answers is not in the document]
bojanrajkovic: No idea how that ended up 4-space indented…when the rest of the file is 2-space…

Agents/Xamarin.Interactive/CodeAnalysis/Evaluating/EvaluationContextId.cs
abock: Cool. When you rebase all this before merge, keep the changes to these to files as a separate commit so we can revert when we bump.

Clients/Xamarin.Interactive.Client.Web/ClientApp/WebAssemblyAgentContainer.ts
```diff
@@ -0,0 +1,146 @@
+import { CodeCellEventType, ICodeCellEvent } from "./evaluation";
+import * as catalog from './i18n'
+import { StatusUIAction, MessageKind, MessageSeverity } from "./messages";
+import { WorkbookSession } from "./WorkbookSession";
+
+export class WebAssemblyAgentContainer
+{
+    private _wasmFrame: HTMLIFrameElement
```
abock: Two things. It took me a second to realize why you were prefixing with `_`.

* In this case, I think you can just make the fields `readonly` since they're both assigned in `constructor`:

    ```typescript
    readonly wasmFrame: HTMLIFrameElement
    readonly wasmMesssageChannel: MessageChannel
    ```

* If however you cannot (and question why, as we should try to be as immutable as possible), keep the field and the property together:

    ```typescript
    private _wasmMessageChannel: MessageChannel
    get wasmMessageChannel(): MessageChannel {
        return this._wasmMessageChannel
    }
    ```

Clients/Xamarin.Interactive.Client.Web/ClientApp/WebAssemblyAgentContainer.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * This is a callback called by the `load` event on {@link _wasmFrame}. It pushes
+     * some events to the upstream session controlling us, actually sets up
+     * the message channel by passing port 2 over to the iframe, and pushes any
+     * queued messages over to the frame so it can execute them when it's done loading.
+     *
+     * @callback
+     */
+    _executeQueuedWasmMessages(): void {
```
abock: Do not prefix with `_` - mark it `private`.

Clients/Xamarin.Interactive.Client.Web/ClientApp/WebAssemblyAgentContainer.ts
```diff
@@ -0,0 +1,146 @@
...
+            this._wasmMessageChannel.port2
+        ]);
+        this.queuedWasmMessages.forEach(msg => this._wasmMessageChannel.port1.postMessage(msg))
+    }
+
+    /**
+     * Dispatches message sent by the WebAssembly agent.
+     * @param ev The message event from the {@link MessageChannel} we opened for communications.
+     */
+    _dispatchWasmMessage(ev: MessageEvent): void {
```
abock: Do not prefix with `_` - mark it `private`.

Clients/Xamarin.Interactive.Client.Web/ClientApp/WebAssemblyAgentContainer.ts
```diff
@@ -0,0 +1,146 @@
+import { CodeCellEventType, ICodeCellEvent } from "./evaluation";
```
abock: Header:

```
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.
```

Clients/Xamarin.Interactive.Client.Web/ClientApp/WebAssemblyAgentContainer.ts
bojanrajkovic: Yep, you're totally right. I forgot about `readonly` semantics in TS/JS. I did realize that `wasmFrameLoaded` does actually need to be a private setter, so it uses the suggested pattern of private field immediately followed by its property. :)

Clients/Xamarin.Interactive.Client.Web/Views/Home/Wasm.cshtml
```diff
@@ -0,0 +1,13 @@
...
+    Layout = null;
+}
+
+<!doctype html>
+<html lang="en-us">
+    <head>
+        <script type="text/javascript" src="/wasm/runtime/main.js"></script>
+    </head>
+    <body>
+        <script type="text/javascript" src="/wasm/runtime/mono.js"></script>
```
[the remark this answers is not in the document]
bojanrajkovic: I want to be absolutely sure main.js loads first—it sets up window-global things that WASM assumes will be there by the time that it is loaded. I do not want to leave any chance that mono.js/WASM is loaded first, and blocking page load by placing main.js in the head is the surest way to do that.

Clients/Xamarin.Interactive.Client.Web/ClientApp/WorkbookSession.ts
```diff
@@ -116,6 +133,18 @@ export class WorkbookSession {
                 targetPlatformIdentifier: description
             }
 
+        // If we've kicked off WASM mode, initialize it, otherwise tear down the frame and message channel
+        // if they were set up.
+        if (description.targetPlatformIdentifier === "webassembly-monowebassembly") {
```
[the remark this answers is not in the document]
bojanrajkovic: Can't hurt.

CodeAnalysis/Xamarin.Interactive.CodeAnalysis/CodeAnalysis/Evaluating/EvaluationService.cs
```diff
@@ -383,6 +383,9 @@ public Task EvaluateAllAsync (CancellationToken cancellationToken = default)
 
                 events.Observers.OnNext (lastCellFinishedEvent);
 
+                if (status != EvaluationStatus.Success)
+                    NotifyEvaluationComplete (evaluatableCodeCell.CodeCellId, status);
```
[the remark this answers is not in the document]
bojanrajkovic: `Success`-ful evals will be pushed into NotifyEvaluationComplete once they've actually been evaluated and represented via the message channel (for normal agents) or via the SignalR hub for WASM. This catches non-Success evals, which otherwise just end up spinning forever.

Clients/Xamarin.Interactive.Client.Web/Startup.cs
```diff
@@ -40,6 +39,7 @@ public void ConfigureServices (IServiceCollection services)
                 WorkbookAppInstallation.RegisterSearchPath (searchPath);
 
             services.AddSingleton (new WebClientAppContainer ());
+            services.AddSingleton (new ReferenceWhitelist ());
```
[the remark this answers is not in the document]
bojanrajkovic: Nah—it would just make it really hard to work with for `AssemblyController`, for little/no benefit.

PR #245, opened 2018-04-27
Exceptional exceptions

abock:
This makes the reflection remoting API immutable and fully JSON serializable. Remove the useless exception filtering we were doing in EvaluationService. We will implement a new exception renderer in TypeScript which can filter out EvaluationContext frames and collapse compiler generated frames.


sandyarmstrong:
🔖

PR #252, opened 2018-05-03
NuGet: move from Client to CodeAnalysis

abock:
Port tests to xUnit.


abock:
1. consumers of XI.CA will be able to get NuGet support (a little more glue work needs to be done, but this is a necessary first step)
2. we did not have an existing XI.Client xUnit project

Issue #249, opened 2018-05-03
Xcode is not correctly detected when opening iOS workbooks


atsushieno:
I have Xcode 9.3 installed but Workbooks.app does not recognize it.

### Steps to Reproduce

1. launch workbooks.app
2. open iOS workbook sample e.g. "learning iOS" sample linked from https://developer.xamarin.com/workbooks/
3. 

### Expected Behavior

It just works.

### Actual Behavior

<img width="773" alt="screen shot 2018-05-03 at 14 19 23" src="https://user-images.githubusercontent.com/53929/39560936-173a35cc-4edd-11e8-93b9-4c707ba12a61.png">

#### Xamarin Workbooks
| Detail | Value |
| :- | :- |
| Version | 1.4.0 |
| Git Branch | master |
| Git Hash | b3f92f9 |
| VSTS Definition | Workbooks & Inspector (Mac) |

#### System Info
| Component | Value |
| :- | :- |
| macOS | Version 10.13.4 (Build 17E202) (64-bit) |
| CPU Cores | 4 / 4 |
| Physical Memory | 16 GB |

#### System-Installed Software
| Component | Version |
| :- | :- |
| Mono | 5.10.1.47 |
| Xamarin Mac | 4.4.1.173 |
| Xamarin iOS | 11.10.1.174 |
| Xamarin Android | 8.3.0-18 |
| Visual Studio | 7.5.0.1222 (705001222) |

### Log File

[Xamarin Workbooks 2018-05-03.log](https://github.com/Microsoft/workbooks/files/1969717/Xamarin.Workbooks.2018-05-03.log)

sandyarmstrong:
Are you able to deploy Xamarin.iOS apps to your Xcode 9.3 simulator from VSmac?

We need to make sure this message gets surfaced better (it's in the log): **Unable to locate a valid Xcode installation. Check the Apple SDK settings in Xamarin Studio preferences or ensure that `xcode-select -p` returns a valid Xcode installation path.**

If you follow these instructions, are you able to get Workbooks (and VSmac, and Xamarin.iOS in general) to recognize it?

sandyarmstrong:
Can you share your `~/Library/Preferences/Xamarin/Settings.plist`?

My guess is that it does now have any xcode path in it, and that VSmac is actually checking the default path first, which we do not do.

[...]

PR #283, opened 2018-05-10
Prefer /Applications/Xcode.app to xcode-select path

sandyarmstrong:
Mimics VSmac behavior.

Fixes https://github.com/Microsoft/workbooks/issues/249.


sandyarmstrong:
This will need a 1.4 backport as well.

sandyarmstrong:
See https://github.com/xamarin/Xamarin.MacDev/blob/master/Xamarin.MacDev/AppleSdkSettings.cs for VSmac behavior. Perhaps we can use Xamarin.MacDev directly in the future.

PR #309, opened 2018-05-21
Fix potential hangs loading iOS workbooks

sandyarmstrong:
1.4 backport of https://github.com/Microsoft/workbooks/pull/313 and https://github.com/Microsoft/workbooks/pull/320.


sandyarmstrong:
Do not merge yet, should probably do master merge first.

Clients/Xamarin.Interactive.MTouch/MTouchSdkTool.cs
```diff
@@ -46,40 +47,80 @@ public static string GetMlaunchPath ()
...
-            throw new Exception ($"'{fileName} {arguments}' exited with exit code {proc.ExitCode}");
+        static async Task<string> RunToolWithRetriesAsync (
+            string fileName,
+            string arguments,
+            int timeoutRetries = 3,
+            int timeout = 5000)
+        {
+            for (var i = 0; i < timeoutRetries; i++) {
+                try {
+                    return await RunToolAsync (fileName, arguments, timeout);
```
abock: Exponential backoff please. This allows you to not care about the actual timeout duration since you compute it from the number of iterations you're willing to attempt. Generally provides a much smoother user experience as well with more attempts within the same overall window.

To fit within the 15 seconds you've got here, you can do 7 attempts with a 100×2ⁱ ms backoff (we use this elsewhere), which would be 12 seconds total wait across 7 attempts:

```csharp
… timeoutRetries = 7 …

for (int i = 0; i < timeoutRetries; i++) {
    try {
        return await RunToolAsync (
            fileName,
            arguments,
            (int)Math.Pow (2, i) * 100));

        …
}
```


backoff

sandyarmstrong: Isn't exponential backoff more about time _between_ retries? The caller here still knows best what kind of timeout is appropriate for the given process (mlaunch can be slow, for example).

So given that we do care about the actual timeout duration, I'm not sure what a backoff approach buys us. This isn't anything like network congestion.

abock: Never mind. It was not clear how this timeout was actually being used. So just disregard the whole bit.

PR #313, opened 2018-05-21
 Fix potential hangs loading iOS workbooks

sandyarmstrong:
On my system, I frequently see `plutil` requests hang for no apparent reason. Fixed process execution in `MTouchSdkTool` to have a proper async flow and logging, and to support timeouts and limited retries. Ideally, `Exec` would be updated to support these features, and `MTouchSdkTool` would just rely on that class. And ideally we would not have to call `plutil` at all. We can address these down the road.

Also fixed a mistake in the `DefaultSdkRoot` path that caused issues for Windows.

### DIFF
--- a/Clients/Xamarin.Interactive.Client.Mac.SimChecker/Program.cs
+++ b/Clients/Xamarin.Interactive.Client.Mac.SimChecker/Program.cs
@@ -9,14 +9,16 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 
+using Xamarin.Interactive.Logging;
 using Xamarin.Interactive.MTouch;
 
 namespace Xamarin.Interactive.Mac.SimChecker
 {
     public class Program
     {
-        public static void Main (string[] args)
+        public static async Task Main (string[] args)
         {
             if (args.Length == 1 && args [0] == "--version") {
                 Console.WriteLine (BuildInfo.VersionString);
@@ -24,9 +26,11 @@ namespace Xamarin.Interactive.Mac.SimChecker
                 return;
             }
 
+            Log.Initialize (new LogProvider (LogLevel.Debug));
+
             string sdkRoot;
             try {
-                sdkRoot = MTouchSdkTool.GetXcodeSdkRoot ();
+                sdkRoot = await MTouchSdkTool.GetXcodeSdkRootAsync ();
             } catch (Exception e) {
                 Console.Error.WriteLine (e.Message);
                 Environment.Exit (100); // Xcode not configured in XS or not installed at /Applications/Xcode.app
@@ -41,7 +45,7 @@ namespace Xamarin.Interactive.Mac.SimChecker
 
             MTouchListSimXml mtouchList;
             try {
-                mtouchList = MTouchSdkTool.MtouchListSimAsync (sdkRoot).Result;
+                mtouchList = await MTouchSdkTool.MtouchListSimAsync (sdkRoot);
             } catch (Exception e) {
                 e = (e as AggregateException)?.InnerExceptions?.FirstOrDefault () ?? e;
                 Console.Error.WriteLine (e.Message);
--- a/Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
+++ b/Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
@@ -1,52 +1,17 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{D495E2AE-BF3E-444B-B3D6-91EF14293831}</ProjectGuid>
+    <TargetFramework>net461</TargetFramework>
     <OutputType>Exe</OutputType>
-    <RootNamespace>Xamarin.Interactive.Mac.SimChecker</RootNamespace>
-    <AssemblyName>Xamarin.Interactive.Mac.SimChecker</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <ProductVersion>8.0.30703</ProductVersion>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <OutputPath>bin\Debug</OutputPath>
-    <DefineConstants>DEBUG;CATALOG_API_ONLY</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ExternalConsole>true</ExternalConsole>
-    <PlatformTarget>x64</PlatformTarget>
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <Optimize>true</Optimize>
-    <OutputPath>bin\Release</OutputPath>
     <DefineConstants>CATALOG_API_ONLY</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ExternalConsole>true</ExternalConsole>
-    <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
+
   <ItemGroup>
-    <Reference Include="System" />
-    <Reference Include="System.Xml.Serialization" />
-    <Reference Include="System.Xml" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\BuildInfo.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\Collections\PropertyList.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\I18N\Catalog.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\Versioning\ReleaseVersion.cs" />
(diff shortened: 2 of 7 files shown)


Clients/Xamarin.Interactive.Client.Mac.SimChecker/Program.cs
```diff
@@ -24,16 +25,18 @@ public static void Main (string[] args)
                 return;
             }
 
+            Log.Initialize (new LogProvider (LogLevel.Info));
+
             string sdkRoot;
             try {
-                sdkRoot = MTouchSdkTool.GetXcodeSdkRoot ();
+                sdkRoot = MTouchSdkTool.GetXcodeSdkRootAsync ().Result;
```
abock: Make the program `async Task Main` and you can `await` these calls.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
```diff
@@ -1,52 +1,31 @@
...
-    <OutputPath>bin\Release</OutputPath>
     <DefineConstants>CATALOG_API_ONLY</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ExternalConsole>true</ExternalConsole>
-    <PlatformTarget>x64</PlatformTarget>
+    <OutputPath>bin\Release</OutputPath>
   </PropertyGroup>
+
   <ItemGroup>
```
abock: You can kill this whole `<ItemGroup>` of references in SDK style projects.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
```diff
@@ -1,52 +1,31 @@
...
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{D495E2AE-BF3E-444B-B3D6-91EF14293831}</ProjectGuid>
+    <TargetFramework>net461</TargetFramework>
     <OutputType>Exe</OutputType>
-    <RootNamespace>Xamarin.Interactive.Mac.SimChecker</RootNamespace>
-    <AssemblyName>Xamarin.Interactive.Mac.SimChecker</AssemblyName>
-    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
-    <ProductVersion>8.0.30703</ProductVersion>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
abock: Take out these properties. Directory.Build.props does the right thing.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
```diff
@@ -1,52 +1,31 @@
...
+    <PlatformTarget>x64</PlatformTarget>
   </PropertyGroup>
+
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <OutputPath>bin\Debug</OutputPath>
     <DefineConstants>DEBUG;CATALOG_API_ONLY</DefineConstants>
```
abock: You do not need to define `DEBUG` anymore. It'll be defined for you by the SDK, as will `NET461`. Move `<DefineConstants>CATALOG_API_ONLY</DefineConstants>` to the only `<PropertyGroup>` this project needs at the top of the project.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
```diff
@@ -1,52 +1,31 @@
...
-    <Compile Include="..\..\Agents\Xamarin.Interactive\Versioning\ReleaseVersion.cs">
-      <Link>ReleaseVersion.cs</Link>
-    </Compile>
+    <Compile Include="..\..\Agents\Xamarin.Interactive\BuildInfo.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\I18N\Catalog.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\Versioning\ReleaseVersion.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\Logging\**\*.cs" />
   </ItemGroup>
   <Import Project="..\Xamarin.Interactive.MTouch\Xamarin.Interactive.MTouch.projitems" Label="Shared" Condition="Exists('..\Xamarin.Interactive.MTouch\Xamarin.Interactive.MTouch.projitems')" />
-  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
```
abock: I'm pretty sure you can reduce the entire project to

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net461</TargetFramework>
    <DefineConstants>CATALOG_API_ONLY</DefineConstants>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="..\..\Agents\Xamarin.Interactive\BuildInfo.cs" />
    <Compile Include="..\..\Agents\Xamarin.Interactive\I18N\Catalog.cs" />
    <Compile Include="..\..\Agents\Xamarin.Interactive\Versioning\ReleaseVersion.cs" />
    <Compile Include="..\..\Agents\Xamarin.Interactive\Logging\**\*.cs" />
  </ItemGroup>
  <Import Project="..\Xamarin.Interactive.MTouch\Xamarin.Interactive.MTouch.projitems" Label="Shared" Condition="Exists('..\Xamarin.Interactive.MTouch\Xamarin.Interactive.MTouch.projitems')" />
</Project>
```

However, I wonder if we should just link `Xamarin.Interactive.dll` at this point since it's not a weird PCL. I also wonder if we can just move the `Xamarin.Interactive.MTouch` into `Xamarin.Interactive.dll` entirely.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
```diff
@@ -1,52 +1,31 @@
...
-    <Compile Include="..\..\Agents\Xamarin.Interactive\I18N\Catalog.cs">
-      <Link>Catalog.cs</Link>
-    </Compile>
-    <Compile Include="..\..\Agents\Xamarin.Interactive\Versioning\ReleaseVersion.cs">
-      <Link>ReleaseVersion.cs</Link>
-    </Compile>
+    <Compile Include="..\..\Agents\Xamarin.Interactive\BuildInfo.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\I18N\Catalog.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\Versioning\ReleaseVersion.cs" />
+    <Compile Include="..\..\Agents\Xamarin.Interactive\Logging\**\*.cs" />
```
abock: Just a note that you might not need to include all this logging if the exec stuff in X.I.MTouch is reverted due to having PropertyList.

sandyarmstrong: Logging was needed anyway, to replace some `Console.Error` writes.

As for whether exec stuff should be reverted, I've thought about it too. We'll still be executing `mlaunch` and `xcode-select`, which weren't causing problems, but I feel like (a) I definitely want timeouts now in case other bits and pieces hang, and (b) retries aren't a bad idea for these sorts of calls in the case of timeouts.

It just feels safer.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
sandyarmstrong: Yeah, this looks better.

We could link it, or we could do multiple SSH copies, or use the existing mechanisms for transferring multiple files via zip+ssh... And I definitely want to get rid of the shared project in master.

Not sure if it's worth putting more time into this right now, though?

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Program.cs
```diff
@@ -9,24 +9,28 @@
...
-        public static void Main (string[] args)
+        public static async Task Main (string[] args)
         {
             if (args.Length == 1 && args [0] == "--version") {
                 Console.WriteLine (BuildInfo.VersionString);
                 Environment.Exit (0);
                 return;
             }
 
+            Log.Initialize (new LogProvider (LogLevel.Info));
```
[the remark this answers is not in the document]
sandyarmstrong: I don't think any of the output will ever get to the user, but yeah, may as well switch to Debug for easier debugging.

Clients/Xamarin.Interactive.Client.Mac.SimChecker/Xamarin.Interactive.Client.Mac.SimChecker.csproj
```diff
@@ -1,52 +1,19 @@
...
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-    <DebugSymbols>true</DebugSymbols>
-    <DebugType>full</DebugType>
-    <Optimize>false</Optimize>
-    <OutputPath>bin\Debug</OutputPath>
-    <DefineConstants>DEBUG;CATALOG_API_ONLY</DefineConstants>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ExternalConsole>true</ExternalConsole>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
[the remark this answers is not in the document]
sandyarmstrong: Why? Don't we want warnings-as-errors? That's false by default. Is the second one you mention `PlatformTarget`? I think there was a reason we wanted this to build as 64-bit and not AnyCPU.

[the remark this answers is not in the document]
sandyarmstrong: I checked, it was x64 because we used to reference XM. We no longer do, so we don't need to specify x64 anymore.

abock:
🦑RE ⚾️

Issue #303, opened 2018-05-15
Special-cased support for ML.NET


abock:
While we still need to figure out a better general solution for NuGet packages that depend on native libraries (since there is no standard/conventional way to know what to load, see #204), we should special-case [ML.NET](https://github.com/dotnet/machinelearning) support just as we do for a number of other interesting NuGet packages that natively integrate with the target platform.

See https://github.com/dotnet/machinelearning/issues/157 for more bug details. We should continue any discussion here.

We are gearing up for a minor bug fix release against the 1.4 series and a fix for ML.NET support should be trivial enough to include.

Thanks @cosmincatalin for filing the original issue.

sandyarmstrong:
One thing to note: ML.NET does not have support for 32-bit platforms, and our .NET Core agent is still 32-bit on Windows. I don't see that changing in 1.4.x.

So this special-cased support will only be for:

* Console (.NET Framework) on 64-bit Windows
* WPF on 64-bit Windows
* Console (Mono) on Mac
* .NET Core on Mac
* Xamarin.Mac

migueldeicaza:
Why can we not do .NET Core agent as 64 on Windows?

abock:
I think we're going to have to eat the size overhead again. I would much rather see ML.NET support on .NET Core 64 bit Windows.

sandyarmstrong:
@abock fine by me. Do you want it for 1.4.1?

abock:
Yes.

sandyarmstrong:
What we ended up doing is adding a 64-bit .NET Core agent app on Windows, and the client picks the correct app depending on the OS.

32-bit Windows still won't be able to use ML.NET, but 64-bit Windows will, with .NET Core or .NET Framework (or the WPF agent app).

On Mac, ML.NET will only be usable on Mono for Workbooks 1.4.1. @bojanrajkovic started work on a way to support native libraries for .NET Core on Mac, but it won't make it for 1.4.1. .NET Core workbooks on Mac continue to have no native library support.

I'm closing this even though the fix is only merged in the 1.4 branch, and not in master.

PR #315, opened 2018-05-22
64-bit .NET Core on Windows

sandyarmstrong:
Preparation for fixing https://github.com/Microsoft/workbooks/issues/303.


Clients/Xamarin.Interactive.Client/WorkbookAppInstallation.cs
```diff
@@ -363,5 +365,28 @@ IEnumerable<string> Pathify (string path)
...
+            var versionsParentDir = path.Substring (0, index);
+            if (!Directory.Exists (versionsParentDir))
+                return path;
+
+            // TODO: Use real rid resolution here, check directory existence, etc
+            string rid;
+            if (ClientApp.SharedInstance.Host.IsMac)
+                rid = "osx-x64";
+            else
+                rid = Environment.Is64BitProcess ? "win-x64" : "win-x86";
```
abock:
```csharp
rid = Runtime.CurrentProcessRuntime.RuntimeIdentifier;
```

Clients/Xamarin.Interactive.Client/WorkbookAppInstallation.cs
```diff
@@ -363,5 +365,28 @@ IEnumerable<string> Pathify (string path)
...
         }
+
+        static string RidPathMapper (string path)
+        {
+            const string ridVar = "{rid}";
+            var index = path.IndexOf (ridVar, StringComparison.Ordinal);
+            if (index < 0)
+                return path;
+
+            var versionsParentDir = path.Substring (0, index);
```
abock: Why do you need to care about this?

Seems to me this entire method could just be hoisted to `pathMappers`:

```csharp
... {
    path => path.Replace ("{rid}", Runtime.CurrentProcessRuntime.RuntimeIdentifier);
}
```

CodeAnalysis/Xamarin.Interactive.CodeAnalysis/CodeAnalysis/Resolving/NativeDependencyResolver.cs
```diff
@@ -31,8 +31,7 @@ internal NativeDependencyResolver (TargetCompilationConfiguration compilationCon
         {
             CompilationConfiguration = compilationConfiguration;
 
-            agentArchitecture = Environment.Is64BitProcess &&
-                (HostEnvironment.OS != HostOS.Windows || compilationConfiguration.Sdk.IsNot (SdkId.ConsoleNetCore))
+            agentArchitecture = Environment.Is64BitProcess
```
abock: This is funny... why is `agentArchitecture` being decided based on the host process? Shouldn't any arch checking just check `CompilationConfiguration.Runtime.Architecture`?

It might also just be best to not touch `NativeDependencyResolver` in this PR at all since that's what @bojanrajkovic is working on.

Clients/Xamarin.Interactive.Client/WorkbookAppInstallation.cs
sandyarmstrong: If you look at the path mapper for the XM workbook apps, it searches in the parent dir for the closest match to the host system. I was thinking we'd do the same here, except with RIDs.

But we're not really doing that, so yeah, a simple replace as you suggest would be fine here.

I'll need to backport your Runtime stuff to 1.4.

CodeAnalysis/Xamarin.Interactive.CodeAnalysis/CodeAnalysis/Resolving/NativeDependencyResolver.cs
sandyarmstrong: I was just removing the checks that were in place because of .NET Core being 32-bit on 64-bit Windows.

But yeah, I expect Bojan is deleting all of this. :-)

abock:
Merging because the macOS Hosted Pool is broken in VSTS right now.

PR #318, opened 2018-05-22
build: force separate Restore and Build targets

abock:
At least provide a better and faster error message when restore isn't
run first.

Ugh. https://github.com/Microsoft/msbuild/issues/2455

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -1,4 +1,4 @@
-<Project InitialTargets="Configure" DefaultTargets="Restore;Build">
+<Project InitialTargets="Configure" DefaultTargets="Build">
   <Import Project="Directory.Build.props"/>
   <Import Project="Directory.Build.targets"/>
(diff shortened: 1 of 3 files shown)


build.proj
```diff
@@ -166,6 +167,14 @@
 
   <!-- Solution Targets -->
 
+  <Target
+    Name="CheckRestore"
+    BeforeTargets="Build">
+    <Error
+      Condition="!Exists('$(RestoreStampFile)')"
+      Text="When building with MSBuild directly, The 'Restore' target must be run first. Run 'msbuild /restore' or 'msbuild /t:Restore'."/>
```
sandyarmstrong: Maybe clarify here that `msbuild /restore` still runs the `Build` target?

abock: Nah. It's pretty obvious.

PR #320, opened 2018-05-23
Hello plist my old friend

abock:
Finally just bring over this old PropertyList implementation from the Banshee days. Freshly updated to be internal, with slightly better immutability, and use some more modern C# constructs.

Ideally we would just depend on larger one from Xamarin.MacDev, but that entire library needs to be split up a bit and actually published on NuGet.

For now this should get us XML and Binary format support without having to shell out to plutil, which is known to be problematic as sometimes it freezes/never exits.


lewing:
I had iThumb flashbacks reading this

PR #331, opened 2018-05-29
Web: Initial IE11 support

sandyarmstrong:


Clients/Xamarin.Interactive.Client.Web/ClientApp/components/CodeCellView.tsx
```diff
@@ -37,7 +37,7 @@ export interface ResultRendererRepresentationMap {
 export interface CodeCellResultRendererState {
     result: CodeCellResult
     representations: ResultRendererRepresentationMap
-    selectedRepresentationKey: string
+    selectedRepresentationKey: string | null
```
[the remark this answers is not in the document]
sandyarmstrong: IIRC, this was not essential, and fixes a bug I also so in Firefox. I forget if I checked Chrome.

I can test later and see if reps work on IE without this change. Happy to drop the commit from this PR.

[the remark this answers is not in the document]
abock: Cool. I'd much rather just drop changes to this file if it's not essential/just-for-console-spam.

Clients/Xamarin.Interactive.Client.Web/ClientApp/components/PackageSearch.tsx
```diff
@@ -6,6 +6,7 @@ import { PrimaryButton } from 'office-ui-fabric-react/lib/Button'
 import { Panel, PanelType } from 'office-ui-fabric-react/lib/Panel'
 import { SearchBox } from 'office-ui-fabric-react/lib/SearchBox'
 import { Icon } from 'office-ui-fabric-react/lib/Icon'
+import 'isomorphic-fetch'
```
abock: Where is this coming from? Were we already pulling it in? Does this only kick in for IE11 or does it replace the native implementation in real browsers?

Clients/Xamarin.Interactive.Client.Web/ClientApp/boot.tsx
```diff
@@ -5,7 +5,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-import './boot.scss'
+import 'es6-shim';
```
abock: Is this shim a no-op for real browsers or does it override native implementations of stuff?

Clients/Xamarin.Interactive.Client.Web/ClientApp/components/PackageSearch.tsx
sandyarmstrong: We were already pulling it in. Although, I'm not really sure why we are using it instead of Github's `whatwg-fetch` (https://github.com/github/fetch), which `isomorphic-fetch` wraps.

Anyway, as I understand it, polyfills like this don't replace anything if the API is already defined. See https://github.com/github/fetch/blob/master/fetch.js#L511.

Clients/Xamarin.Interactive.Client.Web/ClientApp/boot.tsx
sandyarmstrong: It looks like it performs a variety of tests on the implementations of these APIs, because some older browsers have bad implementations. See https://github.com/paulmillr/es6-shim/blob/master/es6-shim.js#L864, for example.

We could import this conditionally.

abock: Nah, I don't mind it if native implementations shine through.

Clients/Xamarin.Interactive.Client.Web/ClientApp/components/PackageSearch.tsx
abock: I would love to know what is pulling it in.

abock: In any case, you should specify it explicitly since we explicitly need it.

abock: Oh, looks like we already had it. @bojanrajkovic?

sandyarmstrong: `isomorphic-fetch` is basically a wrapper that makes sure both the node-side and the client-side get working `fetch` support (hence "isomorphic"). But we don't have a node side. So I think we can just pull in `whatwg-fetch` and have an identical experience with slightly less indirection.

@bojanrajkovic  any comments?

bojanrajkovic: It's coming from draftjs, per `npm ll`:

```
├─┬ draft-js@0.10.5
│ │ A React framework for building text editors.
│ │ git+https://github.com/facebook/draft-js.git
│ │ http://draftjs.org/
│ ├─┬ fbjs@0.8.16
│ │ │ A collection of utility libraries used by other Facebook JS projects
│ │ │ git+https://github.com/facebook/fbjs.git
│ │ │ https://github.com/facebook/fbjs#readme
│ │ ├── core-js@1.2.7
│ │ │   Standard library
│ │ │   git+https://github.com/zloirock/core-js.git
│ │ │   https://github.com/zloirock/core-js#readme
│ │ ├── isomorphic-fetch@2.2.1 deduped
│ │ │   Isomorphic WHATWG Fetch API, for Node & Browserify
│ │ │   git+https://github.com/matthew-andrews/isomorphic-fetch.git
│ │ │   https://github.com/matthew-andrews/isomorphic-fetch/issues
```

bojanrajkovic: We also have it as a dev-dependency, that was added in 4e38514b.

abock: Oh but of course it is coming from `fbjs`. Utility libraries pulling in utilities pulling in utilities pulling in operating systems.  🙄

PR #409, opened 2018-06-13
Telemetry: Use Azure Application Insights

sandyarmstrong:
Simplify our events; move some data to common properties included with
each event.

For now, manually serialize individual events to dictionaries. We can
overhaul this in master.

Use the same approach as dotnet, VS, VScode, etc for getting MAC address
hashes. Using code from `dotnet/cli`.

---

### TODO
* [x] Add some namespacing to event names


sandyarmstrong:
Will update this after the `master` version is reviewed and merged.

sandyarmstrong:
The App Insights SDK is supposed to be non-blocking, but I noticed the dotnet CLI keeps tracking requests on a separate thread (https://github.com/dotnet/cli/blob/v2.1.301/src/dotnet/Telemetry/Telemetry.cs#L73). Building the properties dictionary doesn't seem expensive so I'm not sure what the point is. I can inquire, though.

In any event, I removed our backoff/retry implementation since the SDK should be handling this.

PR #422, opened 2018-06-19
Upgrade to .NET Core 2.1, ASP.NET Core 2.1, SignalR 1.0 ... support C# 7.3 ... and run on Linux

bojanrajkovic:
Upgrade to .NET Core 2.1 across the board. Also fixes support for the server and console client to run on Linux.


sandyarmstrong:
Does this block on a fix for https://github.com/dotnet/coreclr/issues/18269?

bojanrajkovic:
Yes—there's no workaround currently in this branch for it.

abock:
Note: merging anyway because we cannot really do anything about the failure on the macOS VSTS hosted build. The installed Mono on the hosted pool is too old. It'll be updated hopefully soon enough.

Agents/Xamarin.Interactive.Console/Xamarin.Interactive.Console.csproj
```diff
@@ -1,59 +1,12 @@
...
-    <WarningLevel>4</WarningLevel>
-    <ExternalConsole>true</ExternalConsole>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <Prefer32Bit>false</Prefer32Bit>
-  </PropertyGroup>
-  <ItemGroup>
-    <Reference Include="System" />
-  </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="System.ValueTuple">
```
sandyarmstrong: As I understand it, we still need a SVT reference here for to support Windows setups that don't have .NET 4.7.

CodeAnalysis/Xamarin.Interactive.CodeAnalysis.Roslyn/RoslynCompilationWorkspace.cs
```diff
@@ -510,7 +510,7 @@ public void RemoveSubmission (DocumentId documentId, DocumentId nextDocumentId)
             workspace.RemoveProject (project.Id);
         }
 
-        const string assemblyNamePrefix = "🐵🐻";
+        const string assemblyNamePrefix = "🐵BEAR";
```
sandyarmstrong: 😿

Clients/Xamarin.Interactive.Client.Web/Xamarin.Interactive.Client.Web.csproj
```diff
@@ -12,15 +11,7 @@
...
-    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.2" />
-    <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.0.2" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.0.3" />
-    <PackageReference Include="Microsoft.AspNetCore.SpaServices" Version="2.0.3" />
-    <PackageReference Include="Microsoft.AspNetCore.SignalR" Version="1.0.0-preview2-30131" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="2.0.2" />
+    <PackageReference Include="Microsoft.AspNetCore.App" />
```
sandyarmstrong: Why don't we need/want a version here?

abock: Because this package is a new runtime metapackage and is pinned to the .NET Core in use: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/metapackage-app?view=aspnetcore-2.1

Agents/Xamarin.Interactive.Console/Xamarin.Interactive.Console.csproj
abock: This is an SDK style project now, which does in fact copy SVT. By upgrading to SDK style, this is addressed implicitly instead of explicitly. If we can do the same for WPF, we should.

Clients/Xamarin.Interactive.Client.Web/Xamarin.Interactive.Client.Web.csproj
abock: Also FWIW, there _is_ a version behind here. It's just implied. It even will show up as item metadata after MSBuild has evaluated it since it will come in from the SDK. It's actually pretty clever. You can see this because it ends up in our dependencies.json/md.

PR #458, opened 2018-06-25
Telemetry: Use Azure Application Insights

sandyarmstrong:
Simplify our events; move some data to common properties included with each event.

For now, manually serialize individual events to dictionaries. We can overhaul this later.

Use the same approach as dotnet, VS, VScode, etc for getting MAC address hashes. Using code from `dotnet/cli`.


sandyarmstrong:
The App Insights SDK is supposed to be non-blocking, but I noticed the dotnet CLI keeps tracking requests on a separate thread (https://github.com/dotnet/cli/blob/v2.1.301/src/dotnet/Telemetry/Telemetry.cs#L73). Building the properties dictionary doesn't seem expensive so I'm not sure what the point is. I can inquire, though.

In any event, I removed our backoff/retry implementation since the SDK should be handling this.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +76,50 @@ void ObservePreferenceChange (PreferenceChange change)
...
-                } else if (httpClient == null) {
+                    appInsightsClient = null;
+                } else if (appInsightsClient == null) {
                     Log.Info (TAG, "Telemetry will be enabled after restart.");
                 }
             }
         }
 
-        public void Post (ITelemetryEvent evnt)
+        public void Post (Exception exception)
```
abock: Have this take a `LogEntry` instead so you can attach logging properties to it, particularly the message, which can and often does differ from the message _in the exception_.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +76,50 @@ void ObservePreferenceChange (PreferenceChange change)
...
-                "logEvent",
-                new EventObjectStreamContent (this, evnt));
+            if (appInsightsClient == null || !enabled)
+                return;
 
-            if (response.StatusCode == HttpStatusCode.Forbidden) {
-                Log.Info (TAG, "Disabling telemetry at service's direction (403).");
-                DisableTelemetry ();
+            try {
+                var props = new Dictionary<string, string> (commonProperties);
```
abock: `commonProperties` is immutable, so:

```csharp
var props = commonProperties.SetItems (evnt.GetProperties ());
```

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +76,50 @@ void ObservePreferenceChange (PreferenceChange change)
...
+                foreach (var pair in evnt.GetProperties ())
+                    props [pair.Key] = pair.Value;
+
+                appInsightsClient.TrackEvent (
+                    ProducerNamespace + evnt.GetType ().Name,
+                    properties: props);
+                appInsightsClient.Flush ();
```
abock: Is flushing after each event recommended behavior? Is there not a background thread that will process queued events in batches?

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +76,50 @@ void ObservePreferenceChange (PreferenceChange change)
...
+            } catch (Exception e) {
+                Log.Error (TAG, e);
```
abock: You do _not_ want to log exceptions like this from telemetry. This is likely to result in an infinite exception loop of death. There is actually a flag on `LogEntry` that should prevent exceptions from being sent as telemetry, so set that flag?

sandyarmstrong: Cool, I missed that flag.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
sandyarmstrong: Flushing is recommended before the app exits. Flushing on each event was how dotnet CLI implemented it. I can change this.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
sandyarmstrong: Great idea! Especially since the exceptions are somewhat truncated currently in Insights.

Agents/Xamarin.Interactive/Telemetry/Models/AgentSession.cs
```diff
@@ -6,16 +6,30 @@
...
-        public DateTimeOffset Timestamp { get; set; }
         public AgentSessionKind Kind { get; set; }
         public string Flavor { get; set; }
+        public string ExternalTelemetrySessionId { get; set; }
+
+        Dictionary<string, string> ITelemetryEvent.GetProperties ()
+        {
+            var dict = new Dictionary<string, string> ();
+
+            if (AgentSessionId != Guid.Empty)
```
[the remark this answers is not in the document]
sandyarmstrong: `ClientKind` and `AgentFlavor`!

Clients/Xamarin.Interactive.Client/ClientApp.cs
```diff
@@ -161,27 +161,9 @@ protected ClientApp ()
         void PostAppSessionStarted ()
         {
             var appSession = new Telemetry.Models.AppSession {
```
abock: Do we even need a distinct app session event with everything now rolled into the context's global properties?

Clients/Xamarin.Interactive.Client/ClientApp.cs
```diff
@@ -161,27 +161,9 @@ protected ClientApp ()
...
-                Timestamp = DateTimeOffset.UtcNow,
-                Version = BuildInfo.VersionString,
-                BuildHash = BuildInfo.Hash,
-                UpdateChannel = Updater?.UpdateChannel,
-                OperatingSystem = new Telemetry.Models.OperatingSystem {
-                    Version = Host.OSVersion.ToString (),
-                    WordSize = (byte)IntPtr.Size,
-                    CpuWordSize = (byte)(Environment.Is64BitOperatingSystem ? 8 : 4)
-                }
+                UpdateChannel = Updater?.UpdateChannel
```
abock: And this should become another context global property as well now?

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -6,92 +6,69 @@
...
                 enabled = true;
 
-                httpClient = new HttpClient {
-                    BaseAddress = telemetryApiUri
+                appInsightsClient = new TelemetryClient ();
+                appInsightsClient.InstrumentationKey = "@TELEMETRY_INSTRUMENTATION_KEY@";
+                appInsightsClient.Context.Session.Id = sessionid.ToString ();
+                appInsightsClient.Context.Device.OperatingSystem = host.OSName.ToString ();
+
+                var commonProperties = new Dictionary<string, string> {
```
abock: Maybe rename this now? `contextGlobalProperties`?

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -6,92 +6,69 @@
...
-                httpClient.DefaultRequestHeaders.UserAgent.Add (
-                    new ProductInfoHeaderValue (
-                        "XamarinInteractiveTelemetry",
-                        BuildInfo.VersionString));
-            } catch (Exception e) {
-                Log.Error (TAG, "Unable to create HttpClient for telemetry", e);
-            }
-        }
+                foreach (var pair in commonProperties)
+                    appInsightsClient.Context.Properties.Add (pair);
```
abock: `TelemetryContext.Properties` is obsolete (at least in master), you want `TelemetryContext.GlobalProperties` it seems? Eliminate the extra dictionary work by just doing (annoying, I know):

```csharp
var globalProperties = appInsightsClient.Context.GlobalProperties;
globalProperties.Add ("Product Version", BuildInfo.VersionString);
globalProperties.Add (...);
...
```

Remember to add update channel here as well.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +78,68 @@ void ObservePreferenceChange (PreferenceChange change)
...
-            eventsSent++;
+                exceptionTelemetry.Properties.Add ("XIExceptionTag", logEntry.Tag);
+                exceptionTelemetry.Properties.Add ("XIExceptionMessage", logEntry.Message);
+                exceptionTelemetry.Properties.Add ("XIExceptionCallerMemberName", logEntry.CallerMemberName);
+                exceptionTelemetry.Properties.Add ("XIExceptionCallerFilePath", logEntry.CallerFilePath);
+                exceptionTelemetry.Properties.Add ("XIExceptionCallerLineNumber", logEntry.CallerLineNumber.ToString ());
 
-            Log.Verbose (TAG, $"({eventsSent}): {evnt}");
+                appInsightsClient.TrackException (exceptionTelemetry);
+                appInsightsClient.Flush ();
```
abock: 🔖 to remind me of our slack talk

Clients/Xamarin.Interactive.Client/ClientApp.cs
abock: Got it. I think we should allow posting arbitrary event names as strings that don't need to have any specific attached properties, so this sounds like a good first candidate.

I mostly wasn't sure if the act of initializing a new telemetry session would implicitly post an event like this. If we need one we need one.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
sandyarmstrong: `GlobalProperties` is in a newer beta that just came out, will bump to it.

sandyarmstrong: They're going to change the implementation again (https://github.com/Microsoft/ApplicationInsights-dotnet/issues/820#issuecomment-393909974) but I'll still make this change for the sake of clarity.

The linked issue also explains that right now, there's no difference in what goes to the server, but they plan to change that in the future (which will be nice when their UI can differentiate global properties).

sandyarmstrong: `GlobalProperties` release hasn't hit NuGet yet.

Clients/Xamarin.Interactive.Client/ClientApp.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace Xamarin.Interactive
 {
-    abstract class ClientApp
+    abstract class ClientApp : IDisposable
```
abock: `IDisposable` is not the right pattern for this. Just add a plain non-virtual `NotifyProcessWillTerminate` method or something and don't try to be clever with the pattern/`GC.SuppressFinalize`. If we need to make this virtual in the future we can do so.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -6,92 +6,112 @@
...
+                    .SetValue (channelBuffer, 1);
+
+                var config = new TelemetryConfiguration ("@TELEMETRY_INSTRUMENTATION_KEY@", channel);
+
+                appInsightsClient = new TelemetryClient (config);
+
+                appInsightsClient.Context.Session.Id = sessionid.ToString ();
+                appInsightsClient.Context.Device.OperatingSystem = host.OSName.ToString ();
+
+                // TODO: Make these GlobalProperties when we bump to 2.7.0-beta3 or later
```
abock: Please store the dictionary as a local for all of these adds.

1. So you don't have to change 50 of them when you can use GlobalProperties
2. so each Add doesn't have to go through all the property dereferences each time

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +121,87 @@ void ObservePreferenceChange (PreferenceChange change)
...
 
-    sealed class JsonSerializationException : Exception
-    {
-        public JsonSerializationException (string message, Exception innerException)
-            : base (message, innerException)
+        /// <summary>
+        /// Flush queued telemetry events synchronously.
+        /// </summary>
+        /// <param name="timeout">Defaults to 5 seconds.</param>
+        public void Flush (TimeSpan timeout = default)
```
abock:
```csharp
public void Flush (Optional<TimeSpan> timeout = default)
    => channel.Flush (timeout.GetValueOrDefault (defaultFlushTimeout))
```

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
abock: Also please use the same formatting for all properties (each key/value pair on a separate line)

sandyarmstrong: Fixed.

Agents/Xamarin.Interactive/Telemetry/Models/AgentSession.cs
```diff
@@ -6,16 +6,33 @@
...
-        public AgentSessionKind Kind { get; set; }
-        public string Flavor { get; set; }
+        public ClientSessionKind ClientKind { get; set; }
+        public string AgentFlavor { get; set; }
+        public string ExternalTelemetrySessionId { get; set; }
+
+        IEnumerable<KeyValuePair<string, string>> ITelemetryEvent.GetProperties ()
+        {
+            if (AgentSessionId != Guid.Empty)
+                yield return new KeyValuePair<string, string> (nameof (AgentSessionId), AgentSessionId.ToString ());
```
bojanrajkovic: Nit: Wrap this like `ExternalTelemetrySessionId` below?

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +129,88 @@ void ObservePreferenceChange (PreferenceChange change)
...
 
-    sealed class JsonSerializationException : Exception
-    {
-        public JsonSerializationException (string message, Exception innerException)
-            : base (message, innerException)
+        /// <summary>
+        /// Flush queued telemetry events synchronously.
+        /// </summary>
+        /// <param name="timeout">Defaults to 5 seconds.</param>
+        public void Flush (Optional<TimeSpan> timeout = default)
```
abock: One more nit, because I hate methods like this, even though it has a summary doc. Call it `BlockingFlush`?

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -6,92 +6,120 @@
...
         const string TAG = nameof (Telemetry);
-        const string DefaultTelemetryEndpoint = "@TELEMETRY_API_URL@";
+        const string ProducerNamespace = "workbooks/client/";
+
+        static TimeSpan defaultFlushTimeout = TimeSpan.FromSeconds (5);
 
         bool enabled;
-        HttpClient httpClient;
         int eventsSent;
+        TelemetryClient appInsightsClient;
```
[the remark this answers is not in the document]
sandyarmstrong: I set `channel` to `null`, thanks.

Not disposing it to avoid a blocking `Flush` that `Dispose` triggers.

Agents/Xamarin.Interactive/Telemetry/Models/AgentSession.cs
sandyarmstrong: Wrap the lines? Sure, I'll just do it for the whole method.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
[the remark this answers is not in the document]
abock: Does it have a finalizer that calls dispose? I really hope not!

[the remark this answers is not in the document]
sandyarmstrong: It doesn't. The only suspicious Finalizer/Dispose pairing I found was in `DiagnosticsTelemetryModule`, but afaict that doesn't ever get set up unless you are using a default configuration, which we are not.

sandyarmstrong:
I've on tested on Mac so far, but the behavior is as-intended.

The only real problem is the blocking flush potentially hanging the UI for 5 seconds on quit. But without a telemetry channel that persists data to disk, we can't change this unless we want to lose crash reporting.

I seem to be able to reproduce the hang if I follow these exact steps:

1. Launch Workbooks with network on.
2. Wait a second for telemetry to send, then turn network off.
3. Create a workbook, wait for it to connect.
4. Quit Workbooks (with network still off).

Given the other tweaks we made to the channel configuration, the _only_ reason for flushing is to get quit-time events (basically, just crash reporting). So if the hang is unacceptable we should just remove the flush and expect to lose most crash reports until we get a channel that does persistence.

sandyarmstrong:
:+1: on Windows in my testing.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -101,141 +129,95 @@ void ObservePreferenceChange (PreferenceChange change)
...
         {
+            if (channel == null)
+                return;
+
+            var flushTimeout = timeout.GetValueOrDefault (defaultFlushTimeout);
+
+            // Flush can block indefinitely, despite the timeout. This has been
+            // observed when network connectivity is lost during an app run.
+            // Run Flush defensively on a background thread to avoid this.
+            Task.Run (() => channel.Flush (flushTimeout)).Wait (flushTimeout);
```
abock: `channel?.Flush` as `channel` could be come `null` by the time the thread runs.

Clients/Xamarin.Interactive.Client/Telemetry/Client.cs
```diff
@@ -6,92 +6,120 @@
...
+                    "Release Candidate Level Name",
+                    BuildInfo.Version.CandidateLevel.ToString ().ToLowerInvariant ());
+                globalProperties.Add (
+                    "Machine ID",
+                    Sha256Hasher.Hash (MacAddressGetter.GetMacAddress ()));
+                globalProperties.Add (
+                    "Update Channel", updater.UpdateChannel);
             } catch (Exception e) {
-                Log.Error (TAG, e);
+                Log.Error (TAG, "Unable to create AppInsights client for telemetry", e);
```
abock: Any exception/error logging makes me nervous inside of the telem client because we could end up in a weird loop since exceptions will get logged through telemetry. I think we should have a `LogError` local function that always sets the SkipTelemetry flag.

sandyarmstrong: I can add one, sure. But also, I'm going to fix some `enabled` checks so it wouldn't matter.

abock: No real preference, I just don't want to end up in a weird state because we logged something through our normal means.

PR #462, opened 2018-06-27
build: go back to Mono 5.10 on VSTS hosted

abock:
⚠️This is just a test 🚧


abock:
Turns out - it's still bad!

PR #463, opened 2018-06-28
Build: Fix generated solution's configuration mapping

sandyarmstrong:
Projects with "AnyCPU" platform need to be listed as "Any CPU" in the
solution file to match the `SolutionPlatform`.

This gets the solution working correctly in VSmac again.


abock:
🆗

PR #471, opened 2018-07-02
iOS: Prefer iPhone X to iPhone 5s

sandyarmstrong:
Client integrations now have access to all available devices, as well.

This changes SimChecker output to be incompatible with Workbooks < 1.2
(when Windows still required a Workbooks install on the Mac host).

* [x] Test on Mac
* [x] Test on Windows


abock:
As discussed in Slack, we're going to change this to prefer the _last_ device in the simulator list that is an iPhone and 64 bit. This should select the X in Xcode 9/10 while allowing for newer → older device fallbacks.

sandyarmstrong:
Should this backport to 1.4.2 or just 1.5.0?

I could imagine a workbook that made assumptions about the device, but otherwise I don't think it breaks anything.

abock:
I think 1.5 is fine

PR #496, opened 2018-10-04
Bumps: Xamarin, C#, .NET Core

sandyarmstrong:
And increase consistency of previous bumps.


Package/Windows/ClientFiles.wxs
```diff
@@ -465,6 +465,24 @@
       <Component Id="System.Diagnostics.DiagnosticSource.dll" Guid="1E23AA5D-D9D4-4C0C-8A8E-714E5F952CF3">
         <File Id="System.Diagnostics.DiagnosticSource.dll" Source="$(var.Xamarin.Interactive.Client.Windows.TargetDir)\System.Diagnostics.DiagnosticSource.dll" />
       </Component>
+      <Component Id="Humanizer.dll" Guid="5A15CE3C-86F1-4DDA-B1A8-0659D1E0E20F">
```
abock: What's bringing in Humanizer and these SQLitePCLRaw assemblies? Specifically with Humanizer it brings in a ton of localization resources/assemblies as well. I would love to avoid it as a dependency.

Package/Windows/SigningExclusions.json
```diff
@@ -14,6 +14,10 @@
   "Xamarin/Workbooks/Client/MouseKeyboardActivityMonitor.dll": "oss",
   "Xamarin/Workbooks/Client/Newtonsoft.Json.dll": "oss",
   "Xamarin/Workbooks/Client/YamlDotNet.dll": "oss",
+  "Xamarin/Workbooks/Client/SQLitePCLRaw.batteries_green.dll": "oss",
```
abock: Humanizer would need to be excluded as well (though I'd like to drop it if possible...)

sandyarmstrong: If it's not being detected, it must already be signed. I'll verify, though.

Package/Windows/ClientFiles.wxs
sandyarmstrong: Roslyn 2.9. If you like, we can not bump Roslyn. The version we already had in this branch supports 7.3. Roslyn 2.9 is just bugfixes, and I don't actually know if we care about any of them.

Package/Windows/SigningExclusions.json
sandyarmstrong: Yeah, it's already signed.

Package/Windows/ClientFiles.wxs
sandyarmstrong: I assume there will be another Roslyn release that corresponds to VS 15.9 (2.9 corresponds to 15.8), but it might not be available from nuget.org in a timely fashion.

sandyarmstrong: https://github.com/dotnet/roslyn/wiki/NuGet-packages#versioning is always handy, btw.

abock: Yep, if 2.8.2 has 7.3 support then let's stay there.

sandyarmstrong: Done, building now.

global.json
```diff
@@ -1,5 +1,5 @@
 {
   "sdk": {
-    "version": "2.1.301"
+    "version": "2.1.403"
```
abock: [`2.2.100-preview3`](https://www.microsoft.com/net/download/dotnet-core/2.2)?

sandyarmstrong: Honestly I'd prefer to bump to 2.2 in another PR. We did light smoking of 2.1 and 15.9 already.

sandyarmstrong: I'm expecting at least wxs changes and maybe other issues, so it would be nice to get this in while it's green.

sandyarmstrong: Nevermind, I tested locally and it seems fine. Bumped.

PR #509, opened 2018-11-28
Let npm-shrinkwrap get updated

sandyarmstrong:


abock (approved):
Oh, I think I like this version the best!

